Issue #116, opened 2026-05-11
Expose `statistics.py` functions in public `__all__` API


eberrigan:
# Expose `statistics.py` functions in public `__all__` API

**Labels:** `enhancement`, `api`, `priority:high`
**Blocker for:** bloom-mcp MCP tool wrapping; Metcalf 2026 intern Phase 1 (wheat EDPIE reproduction)

## Context

`src/sleap_roots_analyze/statistics.py` contains 8 heritability/ANOVA/variance functions that are not currently exported through the package's `__init__.py`:

| Function | Line |
|---|---|
| `calculate_trait_statistics` | 20 |
| `perform_anova_by_genotype` | 57 |
| `calculate_heritability_estimates` | 159 |
| `identify_high_heritability_traits` | 394 |
| `analyze_heritability_thresholds` | 416 |
| `analyze_trait_variance` | 464 |
| `diagnose_heritability_issues` | 561 |
| `compare_trait_heritabilities` | 707 |

Today `__init__.py` only imports the `cross_experiment_analysis.calculate_genotype_statistics` function — `statistics.py` is missing entirely. As a result, downstream code (including `bloommcp/source/trait_statistics.py`, which is a 810-line **duplicated implementation** of the same functions) imports from internal modules or maintains its own copy.

## What's needed

1. Add a `from sleap_roots_analyze.statistics import (...)` block to `__init__.py` for all 8 functions.
2. Add a `# Statistics / heritability functions` section to `__all__`.
3. Verify each function has a complete docstring (Args, Returns, Raises) and full type hints — these are the source-of-truth for downstream tool schemas.
4. Add module-level docstring to `statistics.py` describing scope (where this fits vs. `cross_experiment_analysis.py`).

## Acceptance

- [ ] All 8 functions importable via `from sleap_roots_analyze import calculate_heritability_estimates` (and the rest).
- [ ] `__all__` lists each by name.
- [ ] `mypy --strict src/sleap_roots_analyze/statistics.py` passes.
- [ ] Sphinx/mkdocs (if used) generates entries for each function.

## Why this is urgent

[...]

PR #140, opened 2026-06-08
Audit __all__ exports for introspection-readiness (#117)

egao28:
Closes #117.

## Summary

Establishes an **introspection-readiness contract** for `sleap_roots_analyze.__all__` and brings every public symbol up to it. The motivation (issue #117) is bloom-mcp's Phase 3 autopop generator, which builds MCP tool descriptors by reading `sleap_roots_analyze.__all__` and, for each name, calling `inspect.signature()` + `typing.get_type_hints()` + `__doc__` to derive parameters, types, defaults, and descriptions. Any public symbol with a missing annotation, an unresolvable type hint, or an unparsable docstring drops out of auto-generation and expands the hand-written fallback surface — directly working against the ≥80% auto-generation target.

This PR fixes every introspection failure **and** makes the bar a permanent, enforced contract so it can't silently regress as the API grows.

**Result: 112/112 `__all__` entries (110 functions + 2 classes) are introspection-ready — was 18 failing.**

## Deliverables (issue #117 acceptance criteria)

- [x] **All `__all__` entries pass the audit** — `python scripts/check_public_api_docs.py` reports 112 passing / 0 failing.
- [x] **CI check committed** — `scripts/check_public_api_docs.py`, exits non-zero on any violation; also enforced on every test run via `tests/test_public_api_docs.py`.
- [x] **Audit report committed** — `docs/public_api_audit_2026.md` (methodology, findings, fixes, result).

Accompanied by the `audit-public-api-introspection` OpenSpec change (proposal, `public-api-introspection` spec, tasks).

## The contract

For each public **function** in `__all__`:
1. Every parameter (excluding `*args`/`**kwargs`) has a type annotation.
2. The return value has a type annotation.
3. `typing.get_type_hints()` resolves without raising.
4. A docstring exists with a `Returns:` section, plus an `Args:` section when the function takes parameters.
5. Every parameter name appears in the docstring body.
6. A `Raises:` section is present when the function body raises a non-trivial exception.

For each public **class** (`VizPipeline`, `VizPipelineConfig`): a class docstring, and every `__init__` parameter beyond `self` annotated and documented.

## Example usage

Run the audit standalone (e.g. as a CI step) — exit code 0 means the public API is introspection-ready:

```console
$ uv run python scripts/check_public_api_docs.py
Public API introspection audit: 112 __all__ entries
  passing: 112
  failing: 0

All public API entries are introspection-ready.
```

Or call it programmatically (this is how the pytest test consumes it):

```python
from check_public_api_docs import run_audit

violations = run_audit()   # list[str]; empty == contract satisfied
assert violations == []
```

When a public function regresses the bar, the report is specific and actionable:

```
  - my_new_function: missing return annotation
  - my_new_function: parameter 'threshold' not documented
  - my_new_function: raises but docstring has no 'Raises:' section
```

## Fixes applied (type hints + docstrings only — no behavior change)

The audit found 18 failing entries. No public signatures or return values were changed.

| Category | Functions | Fix |
|---|---|---|
| `get_type_hints()` → `NameError: 'Any'` | `create_trait_boxplots_by_genotype`, `create_exploratory_summary_plots`, `create_trait_by_genotype_boxplots` | add `from typing import Any` to `visualization.py` |
| Missing return annotation / `Returns:` | `cli.main`, `save_viz_config`, `validate_viz_config`, `create_publication_figure` | add `-> None` and/or a `Returns:` section |
| Raises but no `Raises:` section | 11 functions across `viz_utils`, `data_utils`, `pca`, `clustering`, `outlier_detection`, `visualization` (all `ValueError`; `calculate_optimal_clusters_hierarchical` also `RuntimeError`) | add accurate `Raises:` sections |

The `Any`-import failure is the same class of bug fixed for `statistics.py` in #116: because these modules use `from __future__ import annotations`, annotations are stringized and the unimported-name error only surfaces when `get_type_hints()` evaluates them — i.e. exactly on the bloom-mcp path, not at import time.

## API changes

- **No breaking changes.** Additive type annotations and docstring sections only; all existing call sites and return contracts are unchanged.
- `cli.main` gains a `-> None` return annotation (previously unannotated).
- New module `scripts/check_public_api_docs.py` exposes `run_audit() -> list[str]` for programmatic use.

## Validation

- `python scripts/check_public_api_docs.py` → 112 passing, 0 failing (exit 0)
- `uv run pytest -m "not integration"` → **2011 passed**, 0 failed (2 integration deselected, matching CI)
- `black --check` + `ruff check` clean on all changed files
- `openspec validate audit-public-api-introspection --strict` → valid

## Design notes / reasoning

- **Enforced in pytest, not only as a standalone CI step.** The contract is wrapped by `tests/test_public_api_docs.py` so it runs in the existing test job on all three OSes — no new CI wiring required, and a regression fails the suite immediately. The standalone `scripts/` entry point remains for direct CI use and local runs.
- **`Raises:` detection is deliberately conservative.** The script parses each function's own source via `ast` and only counts `raise SomeError(...)` statements in the top-level body — bare re-raises (`raise`) and raises inside nested functions/lambdas are ignored, avoiding false positives that would push noise into docstrings.
- **Classes are checked via their constructor.** bloom-mcp introspects a class through its `__init__` signature, so the class contract checks the class docstring plus annotated/documented `__init__` params, accepting documentation in either the class or `__init__` docstring.
- **`Raises:` text was derived from the code, not guessed.** Each new `Raises:` entry was written from the function's actual raised exception type and message conditions, so the docs match behavior rather than being boilerplate.
- **Out of scope:** pre-existing `black`/`ruff` violations remain in files this PR does not touch (`scripts/compare_pca_*.py`, `scripts/sanity_check.py`, `scripts/investigate_scanline_first_ind.py`, `tests/test_grouped_pipeline_robustness.py`). They predate this branch and would be a separate cleanup.

### DIFF
--- a/docs/public_api_audit_2026.md
+++ b/docs/public_api_audit_2026.md
@@ -1,0 +1,123 @@
+# Public API Introspection Audit (2026)
+
+**Issue:** [#117](https://github.com/talmolab/sleap-roots-analyze/issues/117) —
+*Audit `__all__` exports for introspection-readiness (type hints + docstrings)*
+
+**Blocker for:** bloom-mcp autopop generator (Phase 3 of the Metcalf 2026 project),
+which targets a ≥80% auto-generation rate for MCP tools.
+
+## Why
+
+bloom-mcp builds MCP tool descriptors by reading `sleap_roots_analyze.__all__` and,
+for each name, calling `inspect.signature()`, `typing.get_type_hints()`, and
+`__doc__` to derive parameters, types, defaults, and descriptions. Any public symbol
+with a missing annotation, an unresolvable type hint, or an unparsable docstring
+drops out of auto-generation and falls back to a hand-written wrapper. This audit
+brings every `__all__` entry up to a machine-introspectable bar and makes that bar a
+permanent, enforced contract.
+
+## Methodology
+
+The audit is implemented as a committed script,
+[`scripts/check_public_api_docs.py`](../scripts/check_public_api_docs.py), and is
+enforced on every test run via
+[`tests/test_public_api_docs.py`](../tests/test_public_api_docs.py). It iterates
+`sleap_roots_analyze.__all__` (112 entries: 110 functions + 2 classes) and applies
+the following criteria.
+
+For each public **function**:
+
+1. Every parameter (excluding `*args`/`**kwargs`) has a type annotation.
+2. The return value has a type annotation.
+3. `typing.get_type_hints()` resolves without raising.
+4. A docstring exists with a `Returns:` section, plus an `Args:` section when the
+   function takes parameters.
+5. Every parameter name appears in the docstring body.
+6. A `Raises:` section is present when the function body raises a non-trivial
+   exception (a `raise SomeError(...)` in its own body — bare re-raises and raises
+   inside nested functions are ignored).
+
+For each public **class** (`VizPipeline`, `VizPipelineConfig`):
+
+7. The class has a docstring.
+8. Every `__init__` parameter beyond `self` is annotated and named in the class or
+   `__init__` docstring.
+
+## Findings (before the change)
+
+18 of the 112 `__all__` entries failed at least one criterion. The 94 others already
+passed.
+
+### Unresolvable type hints (`get_type_hints()` → `NameError: 'Any'`)
+
+`visualization.py` used `Any` in annotations but never imported it. Because the
+module uses `from __future__ import annotations`, the annotations are stringized and
+the failure only surfaces when `get_type_hints()` evaluates them — exactly the
+bloom-mcp path. (Same class of bug fixed for `statistics.py` in #116.)
+
+| Function | Module |
+| --- | --- |
+| `create_trait_boxplots_by_genotype` | `visualization` |
+| `create_exploratory_summary_plots` | `visualization` |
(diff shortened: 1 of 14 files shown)


eberrigan (approved):
Really nice work — textbook TDD: you committed the audit script and a failing test first, then drove it to green, and the `ast`-based `Raises:` detection correctly skips nested functions/lambdas and bare re-raises. CI is green on all three OSes and there's no behavior change. Approving.

One thing to fix before you merge (the code is good — this is about keeping the OpenSpec artifacts honest):

- **Sync the OpenSpec proposal + tasks with what actually shipped.** `proposal.md` and `tasks.md` describe **7** failing functions and only enumerate the `Any`-import + return-annotation fixes. The PR actually fixed **18** — you added `Raises:` sections to 11 more functions, and `spec.md` includes the `Raises:` criterion. The audit report (`public_api_audit_2026.md`) and the PR description already have the right count. Please update `proposal.md` (Why + What Changes) and `tasks.md` so the archived spec of record reflects all 18 / the `Raises:` requirement.

Minor (optional): the "every parameter is documented" check matches the param name anywhere in the docstring (`\bname\b`), so a param mentioned only in `Returns:` would still pass. It's conservative, so fine — a one-line comment noting that limitation would help the next reader.

PR #141, opened 2026-06-09
Audit random_state/seed handling for determinism (#118)

egao28:
Closes #118.

## Summary

Locks in **deterministic reproducibility** of the stochastic public functions so bloom-mcp's Phase 2 golden-value tests (reproducing the wheat EDPIE paper's PCA, clustering, and UMAP results) are stable across runs and CI machines.

**Audit finding:** the implementation was already correct — every genuinely stochastic public function accepts a `random_state` (default `42`) and propagates it to its underlying sklearn/umap estimator, and an empirical two-run check confirms identical output (UMAP embeddings are bit-identical with a fixed seed). The gaps were in the *guarantees*, not the code: there was no regression test pinning this and no documented tolerance policy. This PR adds both. **No production code change was required.**

## Audit results

| Function | Module | `random_state` | Propagated | Deterministic (2-run) |
|---|---|---|---|---|
| `perform_pca_analysis` | `pca` | ✅ (=42) | ✅ via `fit_pca` | ✅ |
| `perform_umap_analysis` | `umap` | ✅ (=42) | ✅ to `umap.UMAP` | ✅ (exact) |
| `perform_kmeans_clustering` | `clustering` | ✅ (=42) | ✅ to `KMeans` | ✅ |
| `perform_gmm_clustering` | `clustering` | ✅ (=42) | ✅ to `GaussianMixture` | ✅ |
| `perform_hierarchical_clustering` | `clustering` | ➖ none | n/a | ✅ (deterministic) |
| `detect_outliers_isolation_forest` | `outlier_detection` | ✅ (=42) | ✅ to `IsolationForest` | ✅ |
| `detect_outliers_kmeans` | `outlier_detection` | ✅ (=42) | ✅ via `perform_kmeans_clustering` | ✅ |
| `detect_outliers_gmm` | `outlier_detection` | ✅ (=42) | ✅ via `perform_gmm_clustering` | ✅ |
| `detect_outliers_mahalanobis` | `outlier_detection` | ✅ (=42) | ✅ via `perform_pca_analysis` | ✅ |

`perform_hierarchical_clustering` uses scipy `linkage`, which is **deterministic** — same input → same linkage matrix. It intentionally takes no `random_state` (a no-op param would misrepresent the API); its determinism is covered by the test instead. No other public function computes a stochastic result without a seed (`create_umap_*` plot helpers consume a precomputed embedding; `create_phenotype_variation_plot` already seeds its jitter).

## Deliverables (issue #118 acceptance)

- [x] Every stochastic public function takes `random_state` and propagates it — verified by audit.
- [x] `tests/test_reproducibility.py` runs each twice and asserts equality.
- [x] Tolerance policy documented in `docs/reproducibility.md`.

## What's added

- **`tests/test_reproducibility.py`** — for each stochastic function: run twice with `random_state=42`, assert **exact** equality for integer labels/outlier indices and **`rtol=1e-6`** for float arrays (embeddings, transformed data, loadings, distances, centers). Plus a `perform_hierarchical_clustering` determinism case and a `random_state=None` smoke test. (17 tests)
- **`docs/reproducibility.md`** — seeding inventory + defaults, the determinism guarantee, the `rtol=1e-6` / exact-labels tolerance policy, and the cross-platform/BLAS caveat (no BLAS pinned; widen comparisons or pin per-test if a golden test proves sensitive).
- OpenSpec change `audit-stochastic-determinism` (proposal, `stochastic-determinism` spec, tasks).

## Design notes

- **Tolerance split:** integer cluster labels / outlier indices are compared exactly; floats use `rtol=1e-6` to leave headroom for cross-platform BLAS reordering. Documented so golden tests adopt the same policy.
- **Hierarchical stays seed-free** rather than gaining a misleading no-op `random_state` — the honest representation of a deterministic algorithm.
- **`random_state` left typed as `int = 42`** (not widened to `Optional[int]`): the functions already forward `None` correctly to sklearn, the smoke test proves it, and golden tests pass a fixed seed — so no signature churn was warranted.

## Validation

- `uv run pytest tests/test_reproducibility.py` → **17 passed**
- `uv run pytest -m "not integration"` → **2027 passed**, 0 failed (2 integration deselected)
- `black --check` + `ruff check` clean on changed files
- `openspec validate audit-stochastic-determinism --strict` → valid

### DIFF
--- a/docs/reproducibility.md
+++ b/docs/reproducibility.md
@@ -1,0 +1,80 @@
+# Reproducibility & Seeding Policy
+
+This document defines how `sleap-roots-analyze` guarantees reproducible results from
+its stochastic analyses, and the numerical tolerance to expect when comparing outputs
+(e.g. for golden-value tests). It backs issue #118 and is enforced by
+[`tests/test_reproducibility.py`](../tests/test_reproducibility.py).
+
+## Seeding policy
+
+Every public function whose result depends on a random number generator accepts a
+`random_state` parameter (default `42`) and forwards it to the underlying
+sklearn / umap estimator. Pass an explicit integer for reproducible output; pass
+`None` to opt into non-deterministic behavior.
+
+| Function | Module | Default `random_state` | Underlying RNG |
+| --- | --- | --- | --- |
+| `perform_pca_analysis` | `pca` | `42` | sklearn `PCA` (matters for the randomized SVD solver) |
+| `perform_umap_analysis` | `umap` | `42` | `umap.UMAP` |
+| `perform_kmeans_clustering` | `clustering` | `42` | sklearn `KMeans` |
+| `perform_gmm_clustering` | `clustering` | `42` | sklearn `GaussianMixture` |
+| `detect_outliers_isolation_forest` | `outlier_detection` | `42` | sklearn `IsolationForest` |
+| `detect_outliers_kmeans` | `outlier_detection` | `42` | via `perform_kmeans_clustering` |
+| `detect_outliers_gmm` | `outlier_detection` | `42` | via `perform_gmm_clustering` |
+| `detect_outliers_mahalanobis` | `outlier_detection` | `42` | via `perform_pca_analysis` |
+
+### Deterministic functions (no seed needed)
+
+`perform_hierarchical_clustering` uses scipy `linkage`, which is **deterministic**:
+the same input always produces the same linkage matrix. It intentionally does **not**
+take a `random_state` — adding one would misrepresent the API. Its determinism is
+still covered by the regression test.
+
+`create_phenotype_variation_plot` adds plotting jitter seeded with
+`np.random.seed(42)`, so its visual output is reproducible. The `create_umap_*` plot
+helpers consume a **precomputed** embedding and run no RNG of their own.
+
+## Determinism guarantee
+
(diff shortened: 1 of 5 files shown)


eberrigan (approved):
Clean, well-reasoned PR — docs + tests only, and the audit correctly concluded no production change was needed. The tolerance policy (exact integer labels, `rtol=1e-6` floats) and the BLAS/cross-platform caveat are thoughtful, and the table-driven `_CASES` parametrization reads well. Approving.

One suggestion worth a fast-follow (does not block merge): the test proves *same seed ⇒ same output*, but a function that ignored `random_state` entirely yet happened to be deterministic would also pass every assertion here. Adding a *different seeds ⇒ different output* check for the genuinely stochastic functions (UMAP, IsolationForest, KMeans) would positively prove the seed is wired through — which is the core of #118. You verified propagation by hand in the audit table; this lets the regression test enforce it.

Minor: `test_accepts_random_state_none` only asserts `isinstance(result, dict)` — fine as a smoke test.

eberrigan (commented):
Hi Evelyn, before mering I would also like you to run some of these functions on test data from the test folders so you understand what is done by each function. You can play around with them in a notebook so you can plot the visualizations. You can also take a look at the pipelines to understand the ordering if the pipelines and how these functions fit in there.

Issue #69, opened 2026-02-17
Bug: run-all does not propagate group_by to viz pipeline — only last QC group gets visualized


eberrigan:
## Summary

When `run-all` executes a QC config with `data.group_by`, the QC pipeline correctly produces one output directory per group. However, the path auto-update mechanism only updates the viz config to point to the **last** QC group's output, causing viz to run only once (for the last group) instead of once per group.

## Expected Behavior (per OpenSpec `add-group-by-timepoint`)

```
run-all manifest.yaml --group-by plant_age_days
```

Should produce:
```
output/
├── qc/
│   ├── plant_age_days_7_<ts>/10_final_data.csv   ✓ (works)
│   ├── plant_age_days_14_<ts>/10_final_data.csv  ✓ (works)
│   └── plant_age_days_21_<ts>/10_final_data.csv  ✓ (works)
└── viz/
    ├── plant_age_days_7_<ts>/                    ✗ (missing)
    ├── plant_age_days_14_<ts>/                   ✗ (missing)
    └── plant_age_days_21_<ts>/                   ✗ (missing)
```

## Actual Behavior

```
output/
├── qc/
│   ├── plant_age_days_7_<ts>/10_final_data.csv   ✓
│   ├── plant_age_days_14_<ts>/10_final_data.csv  ✓
│   └── plant_age_days_21_<ts>/10_final_data.csv  ✓
└── viz/
    └── viz_<pipeline>_<ts>/                      ← only last group (day 21)
```

## Root Cause

`run_all` path auto-update (`qc_mapping` + path injection) only resolves to a single QC output path per viz config. It has no mechanism to:
1. Detect that a QC config produced multiple group outputs
2. Create N viz runs (one per QC group), each pointing to the correct group's `10_final_data.csv`

## What Was Specified But Not Implemented

OpenSpec change `add-group-by-timepoint` tasks 5.1–5.4 (Run-All Integration) and 6.2–6.3 (integration tests) were never completed before PR #63 was merged and the branch closed.

## Workaround

Run viz directly per group:
```bash
# For each QC group output, run viz separately pointing to that group's 10_final_data.csv
sleap-roots-analyze viz viz_config.yaml \
  --data-path "output/qc/plant_age_days_9_<ts>/10_final_data.csv" \
  --output-dir "output/viz/plant_age_days_9/"
```

## Fix Required

[...]

PR #143, opened 2026-06-09
Make columns.replicate optional (#142)

egao28:
## Summary

Makes `columns.replicate` an **optional** config field. Closes #142.

A four-agent code sweep of every consumer (~26 files) confirmed replicate's **values are never used in any computation**: heritability — its only statistical consumer — fits `value ~ 1 + (1|genotype)` and weights by `mean_n_reps` (rows per genotype), neither of which reads the replicate values. Requiring the field is a misfit for datasets with no replicate factor (e.g. Bloom cylinder data, where the plant is the replicate unit). This is the upstream prerequisite for the bloom-mcp analysis-input contract (`talmolab/sleap-roots-contracts#3`).

Removing replicate **changes zero numbers** — the code only *assumed the column was present* at a few access points (which `KeyError`/return-error, not miscompute). This PR gates those points.

## Changes

| Access point | Fix |
|---|---|
| `pipeline/config/utils.py` (`validate_explicit_config`) | Stop rejecting `replicate=None`. Type was already `Optional[str]`. |
| `statistics.py` `calculate_heritability_estimates` | Drop replicate from `required_cols`, the `dropna` subset, and the model-data column rename when `None`. |
| `statistics.py` `analyze_trait_variance` | Accept `replicate_col=None` so public diagnostics (`diagnose_heritability_issues`, `compare_trait_heritabilities`) stay usable on replicate-free data. |
| `pipeline/steps/statistical_analysis.py` | **Not in the original issue.** This in-pipeline heritability consumer hardcoded `replicate_col = "Replicate"`, so with `columns.replicate=None` heritability returned an *error dict* instead of H². Gated to `None` when `columns.replicate` is unset. The issue mislabeled this file as pass-through. |
| `data_cleanup.py` `get_trait_columns` | Already guarded (`if replicate_col:`) — added a regression test. |

Docs: `ColumnConfig.replicate` docstring + `QC_PIPELINE_GUIDE.md` now state replicate is optional and not a model term.

## ⚠️ Not touched: root-core `"Rep"`

The field / root-core pipeline (`aggregate_cores.py`, `merge_all_traits.py` `join_keys`) keys on a **separate, hardcoded `"Rep"`** column from the root-core schema — **not** `columns.replicate`. That `"Rep"` IS load-bearing and is left untouched. A parametrized regression test (`replicate ∈ {None, "rep", "block"}`) proves aggregation is invariant to `columns.replicate`.

## Acceptance criteria

- [x] `columns.replicate=None` (or omitted) passes config validation.
- [x] `calculate_heritability_estimates` runs and returns correct H² with no replicate column (regression test).
- [x] **H² equivalence**: identical H²/variances with vs. without `replicate_col` on the same data — proves values are never load-bearing.
- [x] Cylinder-shaped fixture (genotype → multiple plants, no replicate) runs the QC heritability step and produces real H² (not an error dict).
- [x] Field/root-core path unchanged (parametrized regression test).
- [x] `get_trait_columns` handles `replicate_col=None` without miscounting traits.
- [x] Docs updated.

## Testing

Full suite: **2022 passed** (`uv run pytest`). Black + Ruff clean. OpenSpec change `make-replicate-optional` validates `--strict`.

### DIFF
--- a/configs/templates/qc_cleanup_only_template.yaml
+++ b/configs/templates/qc_cleanup_only_template.yaml
@@ -15,8 +15,12 @@ pipeline_name: "my_cleanup_pipeline"  # REQUIRED: Give your pipeline a descripti
 # ============================================================================
 columns:
   genotype: "geno"           # REQUIRED: Your genotype column name
-  replicate: "rep"           # REQUIRED: Your replicate column name
   barcode: "Barcode"         # Optional: Sample barcode column
+  # replicate is OPTIONAL and disabled by default. Omit this line (or set
+  # `replicate: null`) for data with no replicate factor, e.g. cylinder data
+  # where the plant is the replicate unit. Uncomment and set the column name
+  # only if your dataset actually has a replicate column (issue #142).
+  # replicate: "rep"         # OPTIONAL: Your replicate column name, if present
 
 # ============================================================================
 # REQUIRED: Data Configuration
--- a/configs/templates/qc_full_pipeline_template.yaml
+++ b/configs/templates/qc_full_pipeline_template.yaml
@@ -15,8 +15,12 @@ pipeline_name: "my_full_qc_pipeline"  # REQUIRED: Give your pipeline a descripti
 # ============================================================================
 columns:
   genotype: "geno"           # REQUIRED: Your genotype column name
-  replicate: "rep"           # REQUIRED: Your replicate column name
   barcode: "Barcode"         # Optional: Sample barcode column
+  # replicate is OPTIONAL and disabled by default. Omit this line (or set
+  # `replicate: null`) for data with no replicate factor, e.g. cylinder data
+  # where the plant is the replicate unit. Uncomment and set the column name
+  # only if your dataset actually has a replicate column (issue #142).
+  # replicate: "rep"         # OPTIONAL: Your replicate column name, if present
 
 # ============================================================================
 # REQUIRED: Data Configuration
--- a/configs/templates/qc_template_grouped.yaml
+++ b/configs/templates/qc_template_grouped.yaml
@@ -6,10 +6,11 @@
 # REQUIRED CUSTOMIZATIONS (replace FILL_IN_* placeholders):
 #   - data.csv_path
 #   - data.group_by (column name to group by, e.g., "plant_age_days")
-#   - columns.barcode, columns.genotype, columns.replicate
+#   - columns.barcode, columns.genotype
 #   - pipeline_name
 #
 # OPTIONAL CUSTOMIZATIONS:
+#   - columns.replicate (omit, or set null, for data with no replicate factor)
 #   - cleanup thresholds (max_nan_fraction, max_zeros_per_trait, etc.)
 #   - heritability.threshold (default: 0.30 = permissive)
 #   - outlier_detection settings (default: Mahalanobis chi-squared at 99th percentile)
@@ -42,8 +43,10 @@ columns:
   # Column containing genotype/accession/variety identifiers
   genotype: "FILL_IN_GENOTYPE_COLUMN"  # E.g., "accession_id"
 
-  # Column containing replicate/plant ID (within each genotype)
-  replicate: "FILL_IN_REPLICATE_COLUMN"  # E.g., "plant_id"
+  # OPTIONAL: replicate/plant ID column (within each genotype). Disabled by
+  # default — omit this line (or set `replicate: null`) for data with no
+  # replicate factor, e.g. cylinder data. Uncomment only if your data has one.
+  # replicate: "FILL_IN_REPLICATE_COLUMN"  # E.g., "plant_id"
 
 cleanup:
   # Drop trait columns with >50% zero values (helps remove near-constant traits)
--- a/configs/templates/qc_template_ungrouped.yaml
+++ b/configs/templates/qc_template_ungrouped.yaml
@@ -5,10 +5,11 @@
 #
 # REQUIRED CUSTOMIZATIONS (replace FILL_IN_* placeholders):
 #   - data.csv_path
-#   - columns.barcode, columns.genotype, columns.replicate
+#   - columns.barcode, columns.genotype
 #   - pipeline_name
 #
 # OPTIONAL CUSTOMIZATIONS:
+#   - columns.replicate (omit, or set null, for data with no replicate factor)
 #   - cleanup thresholds (max_nan_fraction, max_zeros_per_trait, etc.)
 #   - heritability.threshold (default: 0.40 = moderate)
 #   - outlier_detection settings (default: Mahalanobis chi-squared at 99th percentile)
@@ -40,8 +41,10 @@ columns:
   # Column containing genotype/accession/variety identifiers
   genotype: "FILL_IN_GENOTYPE_COLUMN"  # E.g., "geno"
 
-  # Column containing replicate/plant ID (within each genotype)
-  replicate: "FILL_IN_REPLICATE_COLUMN"  # E.g., "rep"
+  # OPTIONAL: replicate/plant ID column (within each genotype). Disabled by
+  # default — omit this line (or set `replicate: null`) for data with no
+  # replicate factor, e.g. cylinder data. Uncomment only if your data has one.
+  # replicate: "FILL_IN_REPLICATE_COLUMN"  # E.g., "rep"
 
 cleanup:
   # Drop trait columns with >50% zero values (helps remove near-constant traits)
--- a/docs/QC_PIPELINE_GUIDE.md
+++ b/docs/QC_PIPELINE_GUIDE.md
@@ -57,10 +57,15 @@ You MUST explicitly set these in your config:
 **Column Mappings** (dataset-specific):
 ```yaml
 columns:
-  genotype: "geno"      # Your genotype column name
-  replicate: "rep"      # Your replicate column name
+  genotype: "geno"      # Your genotype column name (required)
+  replicate: "rep"      # Optional — omit or set null if no replicate column
+                        # (e.g. cylinder data). Never used in any computation.
   barcode: "Barcode"    # Sample ID column
 ```
+
+`replicate` is **optional**: its values are never used in any computation and it
+is not a term in the heritability model (`value ~ 1 + (1|genotype)`). Omit it (or
+set `replicate: null`) for datasets with no replicate factor.
 
 **Data Source**:
 ```yaml
--- a/openspec/changes/make-replicate-optional/proposal.md
+++ b/openspec/changes/make-replicate-optional/proposal.md
@@ -50,5 +50,12 @@ load-bearing for field aggregation/merge and must remain untouched.
 - Affected specs: `config-management` (new requirement: Optional Replicate Column).
 - Affected code: `pipeline/config/utils.py`, `statistics.py`, `data_cleanup.py`
   (test only), `pipeline/config/components.py` (docstring), config example docs.
-- Backward compatible: existing configs that set `replicate` keep working
-  identically; the field simply becomes omittable.
+- Mostly backward compatible: existing configs that set `replicate` explicitly
+  keep working identically. **Default-behavior change:** `ColumnConfig.replicate`
+  now defaults to `None` (was `"rep"`), so omitting the key disables replicate
+  instead of silently assuming a `"rep"` column. No shipped config relies on the
+  old implicit default (every replicate-bearing config sets it explicitly, and
+  `configs/qc_cylinder_edpie.yaml` already omits it); a stray `rep` column under
+  the old default would only have been reclassified as a trait, never miscomputed.
+- The four golden QC templates (`configs/templates/qc_*`) now present `replicate`
+  as optional (commented out / omittable) rather than a REQUIRED field.
--- a/openspec/changes/make-replicate-optional/proposal.md
+++ b/openspec/changes/make-replicate-optional/proposal.md
@@ -1,0 +1,54 @@
+# Make columns.replicate optional
+
+## Why
+
+`columns.replicate` is currently a **required** config field, but its values are
+**never used in any computation**. A four-agent adversarial sweep of every
+consumer (~26 files) found no `groupby`/`pivot`/aggregation/dedup/merge keyed on
+the config replicate column. Heritability — its only statistical consumer — fits
+`value ~ 1 + (1|genotype)` (`statistics.py`) and weights by `mean_n_reps`
+(rows per genotype), neither of which reads the replicate values.
+
+Requiring the field is a misfit for datasets with no replicate factor — e.g.
+Bloom cylinder data, where the plant/`barcode` is the replicate unit and there is
+no separate block column. Making it optional is the upstream prerequisite for the
+bloom-mcp **analysis-input contract** (`talmolab/sleap-roots-contracts#3`), which
+needs `replicate` genuinely optional. Removing it **changes zero numbers**; the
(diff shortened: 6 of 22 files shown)


eberrigan (changes requested):
## Review Summary

The `replicate: null` (explicit) path is correct, well-targeted, and backed by a clean equivalence test that genuinely proves replicate values are never load-bearing in `value ~ 1 + (1|genotype)`. The mechanical refactor in `statistics.py` is solid. However, the PR's **headline affordance — "omit `replicate`" — does not work and crashes the pipeline**, because the dataclass default was never changed. Three independent review lenses plus direct code tracing converged on this. Requesting changes for the blocking item; the rest are tighten-ups.

## Blocking Issues

**B1 — The documented "omit `replicate`" path crashes the pipeline (does not behave like `null`).**
`ColumnConfig.replicate` still defaults to `"rep"` (`pipeline/config/components.py:114`) — this PR did not change it. Config loading is `OmegaConf.merge(structured, yaml)`, so an **omitted** field takes the dataclass default:

| YAML | `config.columns.replicate` | Result on replicate-free (cylinder) data |
|---|---|---|
| omit `replicate:` (per the new docs) | `"rep"` | **CRASH** |
| `replicate: null` | `None` | OK — real H² |
| `replicate: ""` | `""` | **CRASH** |

Crash chain (traced in source, not hypothetical):
1. `"rep" is not None` → `statistical_analysis.py:90` sets `replicate_col = "Replicate"`.
2. `calculate_heritability_estimates(..., replicate_col="Replicate")` — column absent → returns a **top-level** `{"error": "Missing required columns: ['Replicate']"}`.
3. The heritability record loop (`statistical_analysis.py:170-200`), unlike the sibling ANOVA loop (line 105), has **no `isinstance(result, str)` guard**. Iterating a top-level error dict yields `trait="error"`, `result="<message string>"`; `"error" in result` is a substring miss → `else` branch calls `result.get(...)` on a `str` → **`AttributeError`**, pipeline dies.

A cylinder user who follows the PR's own docs ("Omit it … for datasets with no replicate factor", `QC_PIPELINE_GUIDE.md`, `ColumnConfig.replicate` docstring) gets a hard crash. This is Copilot comment #1, unaddressed.

Fix — do **both**:
- Make omit actually mean "no replicate": change the config-field default to `None` (or `MISSING`), **or** correct every doc/spec to say replicate-free datasets MUST set `replicate: null` and drop the "omit" claim. The spec scenario currently conflates "validation passes" with "works".
- Add the `isinstance(result, str)` guard to the heritability loop (`statistical_analysis.py:170-200`) so a top-level error dict degrades to an error row instead of crashing — matching the ANOVA loop.

## Important Issues

**I1 — `replicate: ""` is interpreted inconsistently (truthy vs `is not None`).** `statistical_analysis.py:90` uses `is not None` (so `""` → looks for "Replicate" → crash via B1), but `load_trait_data` / `get_trait_columns` use `if replicate_col:` (so `""` → treated as unset). Same config value, opposite meaning across layers. Normalize empty-string → `None` at config validation, or use one consistent truthy convention everywhere. (Copilot #4.)

**I2 — Equivalence is NOT unconditional: it breaks when the replicate column has NaNs.** The `dropna()` subset is `df[[trait, genotype]]` when `replicate_col=None` but `df[[trait, genotype, replicate]]` when set. If the replicate column has any NaN rows, the two subsets differ → different `len(subset)` (can flip the `<4` gate), different `mean_n_reps`, different `n_observations`, and therefore **different H²** between `replicate: "rep"` and `replicate: null` on the same data. The proposal/spec/test docstring assert "identical results" unconditionally — false in general. Preferred fix: **drop `replicate_col` from the `dropna()` subset even when present** (`statistics.py:285`, `:554`) — it is never a model input, so this makes the two paths byte-identical *and* fixes a latent bug where NaNs in an unused metadata column silently shrink the heritability sample. Add a NaN-injection regression test.

**I3 — `analyze_trait_variance` raises a raw pandas `KeyError` for a non-None-but-absent `replicate_col`.** Unlike `calculate_heritability_estimates` (which checks `missing_cols` and returns a structured error), `analyze_trait_variance` does `df[subset_cols].dropna()` with no guard (`statistics.py:~562`). Public diagnostics path stays brittle. (Copilot #3, unaddressed.)

**I4 — Second ungated consumer: `FilterHeritabilityStep` hardcodes `replicate_col="Replicate"`.** `filter_heritability.py:233` calls `compare_trait_heritabilities(..., replicate_col="Replicate")` (reachable when `heritability.generate_diagnostics=true` and traits are removed). On the *working* `replicate: null` path this raises a `KeyError`, but it's swallowed by the broad `except Exception` at line 322 → diagnostics are **silently skipped** (warning logged, no CSV/plots). Not a crash, but replicate-free users requesting diagnostics get none, with no clear signal. Gate this on `config.columns.replicate` like `statistical_analysis.py`. (Task 4.2 fixed `statistical_analysis.py` but missed this structurally identical consumer — worth re-sweeping.)

**I5 — Missing tests for the user-facing paths.** Every new test passes `replicate=None`/`replicate_col=None` *explicitly*; nothing covers: (a) a YAML that **omits** `replicate` round-tripped through `load_qc_config` (the documented path — and the one that's broken); (b) `replicate: ""`; (c) a NaN-bearing replicate column (I2); (d) a **full pipeline** run (load CSV → cleanup → stats) on replicate-free data — the cylinder test exercises only `StatisticalAnalysisStep` in isolation with pre-sanitized data and hand-built metadata, so `load_trait_data` and `CleanupTraitsStep` are never exercised without a replicate column.

## Suggestions

- The equivalence test (`test_heritability_replicate_none_equivalent_to_present`) is a valid regression guard but its docstring oversells it: with the model reading only genotype+trait, the result is identical for *any* replicate values (even a column scrambled to correlate with the trait), so it cannot exhibit a case "where replicate could have mattered". Inject a scrambled/NaN-bearing rep column to make it a meaningful proof.
- `compare_trait_heritabilities` and `diagnose_heritability_issues` are still typed `replicate_col: str = "rep"`; widen to `Optional[str]` to match the functions they now call.
- The proposal claims gating on "≥2 rows per genotype", but no such guard exists — a single genotype still returns a meaningless H² with `n_genotypes=1`. Worth a guard or at least a pinned test on the now-reachable cylinder edge case.
- `tasks.md` item 7.2 (`/lint` clean; full pytest green) is left unchecked `[ ]` though the PR body claims 2022 passed — minor bookkeeping.

## What the PR gets right

- `replicate: null` heritability path is correct; with-replicate path is byte-identical, so no published results change for existing configs.
- Field/root-core `"Rep"` correctly identified as separate and left untouched; the parametrized invariance test (`None`/`"rep"`/`"block"`) is a good guard.
- `get_trait_columns` / `load_trait_data` already `None`/`""`-safe.
- CI green on all three OSes; no performance, memory, or cross-platform concerns (changes are O(1) conditional list construction).

---

eberrigan (commented):
> **Verdict: COMMENT** — re-review of `609c8f16`

## Strong follow-up — almost everything addressed

The fix commit resolves the blocking crash and nearly every important item, with good regression tests for each. Confirmed addressed:

| Item | Status |
|---|---|
| **I2 — NaN-in-replicate equivalence** | ✅ Fixed the right way: replicate is now dropped from the `dropna()` subset entirely (`statistics.py` `subset = df[[trait, genotype_col]].dropna()`), so present-but-NaN replicate yields byte-identical H² to `None`. Test `test_heritability_replicate_with_nans_identical_to_none`. |
| **I1 / Copilot #4 — `""` vs `is not None`** | ✅ Truthy `if config.columns.replicate` everywhere (`statistical_analysis.py`, `cleanup_traits.py`, `filter_heritability.py`, `statistics.py`). Test `test_replicate_empty_string_treated_as_unset`. |
| **I3 / Copilot #3 — `analyze_trait_variance` KeyError** | ✅ Now returns a structured `{"error": ...}` for a named-but-absent column. Test `test_analyze_trait_variance_missing_replicate_returns_structured_error`. |
| **I4 — `FilterHeritabilityStep` hardcoded "Replicate"** | ✅ Gated; plus a bonus `_has_computed_h2` guard so uncomputable-H² traits are retained instead of cascading to a trait-less DataFrame, plus `mean_h2 is None` handling in `generate_summary_viz`. Tests #6 and #7b. |
| **Single-genotype degenerate H²** | ✅ New `len(reps_per_geno) < 2` guard returns a structured error instead of a meaningless H². Test `test_heritability_single_genotype_returns_error`. |
| **`Optional[str]` signatures** | ✅ `diagnose_heritability_issues` and `compare_trait_heritabilities` widened. |
| **Copilot #1 — omit vs "rep" default** | ✅ at the code-docstring level — `ColumnConfig.replicate` now clearly states omitting keeps `"rep"` and does NOT disable; explicit `null`/`""` disables. |

## Remaining items

**1. (Important — unaddressed Copilot #2) `docs/QC_PIPELINE_GUIDE.md` still tells users to "omit it", contradicting the corrected docstring.** The guide at HEAD still reads:
```yaml
  replicate: "rep"      # Optional — omit or set null if no replicate column
```
> `replicate` is **optional**: ... Omit it (or set `replicate: null`) for datasets with no replicate factor.

But the `ColumnConfig.replicate` docstring you just fixed says the opposite — omitting keeps `"rep"` and does **not** disable replicate. A cylinder user who follows the guide ("omit it") gets `replicate="rep"`, and on data with no `rep` column the pipeline fails. The guide needs the same correction as the docstring: replicate-free datasets MUST set `replicate: null` (or `""`); omitting keeps the `"rep"` default. This is Copilot's `QC_PIPELINE_GUIDE.md:68` thread, still unresolved/not-outdated.

**2. (Important) The one true end-to-end test isn't run in CI.** `test_qc_pipeline_runs_on_csv_with_no_replicate_column` — the only test exercising the real user path (CSV → `load_trait_data` → `CleanupTraitsStep` → stats with no replicate) — is decorated `@pytest.mark.integration`, and CI runs `pytest -m "not integration"` (#69), so it's skipped. The other 6 tests in the file run and are great, but the load+cleanup path that motivated this test is unguarded against regression. Consider dropping the marker (it's a small in-memory run) or adding a CI job that runs it.

**3. (Minor — defensive) The heritability record loop is still unguarded.** `statistical_analysis.py:175` does `if "error" in result:` then `result.get(...)` in the `else`. If `calculate_heritability_estimates` ever returns a *top-level* `{"error": ...}` (e.g. the documented-unsupported omit-on-cylinder case), iterating it yields `result="<message string>"`, the substring check misses, and `.get()` raises `AttributeError` instead of writing error rows. The sibling ANOVA loop (line 105) already guards with `isinstance(result, str)`; mirroring it here would make the failure mode a clean error row rather than an opaque traceback.

Items 1 and 2 are worth doing before merge; item 1 in particular is a user-facing trap and closes the last open Copilot thread. Nice work on the rest.

---

eberrigan (commented):
> **Verdict: COMMENT** — re-review of `9590f151`

## The default-to-`None` change is the right call

Changing `ColumnConfig.replicate` to default `None` resolves the omit-vs-null trap at the root rather than papering over it with docs — now omitting genuinely disables replicate, so the `QC_PIPELINE_GUIDE.md` "omit it … for datasets with no replicate factor" wording is finally *accurate*, and Copilot threads #1 (`components.py`) and #2 (`QC_PIPELINE_GUIDE.md`) are both closed at the source. CI green on all three OSes.

Nice side effect: the shipped `configs/qc_cylinder_edpie.yaml` already **omits** `replicate` (only sets `barcode`/`genotype`), so it would have silently defaulted to a nonexistent `"rep"` before — this change makes that real config correct. I also confirmed every replicate-bearing shipped config (`alfalfa_gwas_*`, `emily_shane_*`, `root_core_replicated`, …) sets `replicate` **explicitly**, so the default change doesn't alter any of them.

## Item 1 (docs) — ✅ done (via the behavior fix). Items 2 and 3 from the last round are still open:

**Item 2 — end-to-end test still skipped in CI.** `test_qc_pipeline_runs_on_csv_with_no_replicate_column` is still `@pytest.mark.integration` (line 83), so CI (`-m "not integration"`) doesn't run it, and no fast load→cleanup→stats chained test was added. The load+cleanup path remains un-gated against regression. (Suggested chained-step test from the previous comment still applies.)

**Item 3 — heritability record loop still unguarded.** `statistical_analysis.py:176` is still `if "error" in result:` with no `isinstance(result, str)` guard, so a top-level `{"error": ...}` would still raise `AttributeError` rather than producing error rows. (One-line fix mirroring the ANOVA loop.)

## New finding from this commit (the default change has a doc ripple)

**Important — golden templates still say `replicate` is REQUIRED, contradicting the new default.** The PR changed the schema but didn't update the templates users actually scaffold from:

- `configs/templates/qc_full_pipeline_template.yaml:18` → `replicate: "rep"  # REQUIRED: Your replicate column name`
- `configs/templates/qc_cleanup_only_template.yaml:18` → same `# REQUIRED`
- `configs/templates/qc_template_grouped.yaml` / `qc_template_ungrouped.yaml` → list `columns.replicate` under "REQUIRED CUSTOMIZATIONS (replace FILL_IN_* placeholders)" with `replicate: "FILL_IN_REPLICATE_COLUMN"`

A cylinder user scaffolding from these is told replicate is mandatory — the exact friction this PR set out to remove. Please update the four templates to present `replicate` as optional (omittable / `null` for replicate-free data), and drop it from the "required fields you MUST set" header comments. This is the "config schema change → update golden templates" item.

**Minor — proposal/description understates the change.** The proposal says "Backward compatible: existing configs that set replicate keep working identically; the field simply becomes omittable." True for explicit configs, but the *default itself changed* from `"rep"` to `None`. Harmless in practice (no shipped config relies on the implicit default, and a stray `rep` column would only be reclassified as a trait — never miscomputed), but worth a one-line note in the proposal / CHANGELOG since it's technically a default-behavior change.

---

Summary: the code is now correct and the central design question is settled well. Remaining before merge: **update the 4 golden templates** (important), plus the still-open fast-test (item 2) and loop guard (item 3). The templates one is the only genuinely user-facing gap left.

---

eberrigan (approved):
## Review Summary

All review items across every round are now resolved, and CI is green on Lint + Tests (Ubuntu / Windows / macOS, Python 3.11) at `9a7a430a`. Approving.

The PR cleanly delivers optional `columns.replicate`: replicate values are confirmed never load-bearing in `value ~ 1 + (1|genotype)`, the default now sensibly defaults to `None` (omitting disables it, matching the docs), and the field/root-core `"Rep"` column is correctly left untouched.

## Resolved over the review

- **Default-to-`None`** — omitting `replicate` now genuinely disables it; closes Copilot threads on `components.py` and `QC_PIPELINE_GUIDE.md` at the root. Also fixes the shipped `qc_cylinder_edpie.yaml` (which omits replicate).
- **NaN-equivalence** — replicate dropped from the `dropna()` subset entirely, so present-but-NaN replicate yields byte-identical H² to `None` (regression test included).
- **Consistency** — truthy `if config.columns.replicate` everywhere; `""` handled like `None`; `analyze_trait_variance` returns a structured error instead of a raw `KeyError`; `FilterHeritabilityStep` gated (+ retains uncomputable-H² traits); single-genotype guard.
- **Loop guard** — heritability record loop now guards the string-result case (mirrors the ANOVA loop), with a regression test for the top-level-error path.
- **Tests** — fast `LoadDataStep → CleanupTraitsStep → StatisticalAnalysisStep` chain runs in CI (the prior end-to-end test was `@pytest.mark.integration` and skipped).
- **Templates & docs** — all four golden QC templates present `replicate` as optional; proposal documents the default-behavior change.

Nice, disciplined iteration — each round added a pinning test alongside the fix. LGTM.

---

Issue #130, opened 2026-06-04
Epic: serializable dataclass result types for analytical functions


eberrigan:
**Labels:** `epic`, `enhancement`, `api`, `priority:high`
**Related:** #116, #117, #118, #120, bloom-mcp tool surface (Metcalf 2026)

## Context

Most public analytical functions in `sleap_roots_analyze` return a `Dict` (or tuple) mixing JSON-hostile members: `numpy.ndarray`, sklearn objects (`PCA`, `StandardScaler`, `UMAP`), and `pandas.DataFrame`. A survey of `__all__` found that the large majority of analytical/viz functions cannot cross a JSON boundary without bespoke conversion. This blocks:

- exposing functions as MCP tools (bloom-mcp, Metcalf 2026),
- caching/serializing results as artifacts,
- stable, schema-checkable outputs (FAIR: interoperable + reusable).

We already have a clean house pattern in [`summary/cross_platform_summary.py`](../../src/sleap_roots_analyze/summary/cross_platform_summary.py): flat **stdlib `@dataclass`** result types (`TraitReductionStats`, `PowerStats`, `CorrelationStats`, `TopCorrelation`, …) with typed fields and `@property` derivations. This epic extends that pattern to the analytical functions, starting with the ones bloom-mcp exposes first.

## Principle

- **stdlib `dataclasses`**, not Pydantic — the package stays dependency-light and MCP-agnostic; bloom-mcp wraps these plain dicts in its own Pydantic models at its boundary.
- Result objects hold **only serializable science**; sklearn objects stay out of the clean view (still available via the legacy dict for in-process callers).
- Every field JSON-safe ⇒ `dataclasses.asdict(result)` round-trips with no custom serializer.

## Backwards-compat (applies to every child)

**Additive only.** Existing callers (`result["loadings"]`, the wheat EDPIE paper, the pipeline) must keep working.
1. Add the dataclass + a `from_*_dict()` adapter; functions keep returning the dict → MINOR bump.
2. Later (separate, deprecation-windowed) optionally return the dataclass with a `__getitem__` shim → MAJOR bump.

## Implementation (test-first)

For each child: write the `json.dumps(asdict(...))` round-trip test (and golden/determinism test where relevant) FIRST, watch it fail, then add the dataclass. This is the pairing loop and it feeds the reproducibility CI gate.

## Children

- [ ] #127 — `PCAResult` (the detailed exemplar)
[...]

PR #148, opened 2026-06-11
Add PC-correlation & trait-enrichment workflows (#119)

egao28:
Implements #119. Supersedes the closed #145. Fresh, spec-driven implementation following the `/new-feature` flow.

## Summary
Ports the wheat-EDPIE paper's two cross-platform analyses into the public API as DAG-structured, reusable workflows, then applies the architecture review to give each its natural home.

### Two public workflows
- `cross_platform_pc_correlations(pipeline_run, platform_config, output_dir, …) -> CrossPlatformPCResult` — PC-level, all-platforms synthesis: load per-platform PCA outputs + QC genotypes → aggregate sample PC scores to genotype means → align on common genotypes → every-PC×every-PC correlations per pair → **combined + per-pair FDR** → CSVs/figures/metadata. Returns a **typed `CrossPlatformPCResult`** (result-types pattern, #127–#130) and validates its own `fdr_methods` (full statsmodels family incl. `bonferroni`).
- `trait_correlation_enrichment(correlation_files, output_dir, …) -> dict` — ad-hoc / combined-synthesis binomial enrichment over existing `cross_platform_correlations.csv` (typed `EnrichmentResult`s).

### Trait enrichment as a config-gated DAG step
`CalculateTraitEnrichmentStep` is wired into the existing cross-platform pipeline (step 04, visualize→05), **default-off** via `CrossPlatformConfig.enrichment_enabled`, so it ships automatically under `cross-platform` + `run-all` without changing existing runs. Per-pair, **nominal-p exact binomial, no FDR**, pass-through. `enrichment_p_value_column` is validated to match `correlation_method`.

### Also
- New `pc-correlations` CLI subcommand; reproduction scripts call the public functions (no hard-coded paths).
- CI/power/min-detectable helpers **reused** from `cross_experiment_analysis` (single source of truth).
- Docs: `CROSS_PLATFORM_ANALYSIS.md` workflow section; CHANGELOG entry.

## Architecture decisions (see `openspec/.../design.md`)
- Orchestration shape (disk-in, artifacts-out) supersedes #119's in-memory sketch — matches the real paper code + golden outputs while preserving sample-PCA → genotype-mean → correlate ordering.
- Two **independent** capabilities (PC tests ≠ trait tests); enrichment is not downstream of the PC workflow.

## Verification
- **Full suite: 2071 passed, 1 skipped, 0 failed.**
- black + ruff (`src`) clean; pydocstyle clean; public-API introspection guard 116/116.
- `openspec validate add-pc-correlation-workflows --strict` ✅; tasks checked off.
- Pre-PR subagent self-review: no BLOCKING issues; flagged nits applied.

## Acceptance (#119)
- [x] Functions + `CrossPlatformPCResult`/`EnrichmentResult` exported in `__all__`
- [x] Full type hints + Google docstrings
- [x] Synthetic-fixture unit tests proving each DAG through outputs + workflow independence
- [x] CLI/reproduction entry points call the public functions
- [ ] Wheat-EDPIE golden regression (47/19/0) — known-pending placeholder; activates when #120 ships all 3 platforms' PCA slices (only turface_19 today)

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -8,6 +8,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- Public cross-platform PC-correlation and trait-enrichment workflows (#119).
+  Two new top-level functions wrap the wheat-EDPIE analyses as single calls:
+  `cross_platform_pc_correlations` (PC-level; loads per-platform PCA outputs,
+  aggregates to genotype-mean PC scores, correlates every PC across platform
+  pairs with combined/per-pair FDR; returns a typed `CrossPlatformPCResult`) and
+  `trait_correlation_enrichment` (binomial enrichment over existing
+  `cross_platform_correlations.csv`, returning typed `EnrichmentResult`s). Both
+  exported in `__all__`. A new `pc-correlations` CLI subcommand drives the PC
+  workflow.
+- Optional, config-gated trait-enrichment step in the cross-platform pipeline
+  (`CrossPlatformConfig.enrichment_enabled`, default `False`), with
+  `enrichment_p_value_column` validated against `correlation_method`. When
+  enabled it runs a per-pair exact binomial test (nominal p, no FDR) and writes
+  `trait_enrichment.csv`; existing runs are unchanged.
 - Expose the eight `statistics.py` functions through the top-level
   `sleap_roots_analyze` namespace and `__all__`, so they can be imported directly
   (e.g. `from sleap_roots_analyze import calculate_heritability_estimates`) instead
--- a/docs/CROSS_PLATFORM_ANALYSIS.md
+++ b/docs/CROSS_PLATFORM_ANALYSIS.md
@@ -865,7 +865,7 @@ result = cross_platform_pc_correlations(
     primary_fdr_method="fdr_by",
     fdr_scope="both",
 )
-result["summary"]  # {"n_tests": 47, "n_genotypes": 19, "n_significant_combined": 0, ...}
+result.summary  # {"n_tests": 47, "n_genotypes": 19, "n_significant_combined": 0, ...}
 ```
 
 CLI reproduction: `uv run scripts/run_pc_correlations.py --pipeline-run <dir> --output-dir <dir>`.
--- a/docs/CROSS_PLATFORM_ANALYSIS.md
+++ b/docs/CROSS_PLATFORM_ANALYSIS.md
@@ -14,6 +14,7 @@ This guide covers the cross-platform correlation analysis pipeline, which compar
 - [Output Files](#output-files)
 - [Interpreting Results](#interpreting-results)
 - [Examples](#examples)
+- [Public PC-Correlation and Trait-Enrichment Workflows](#public-pc-correlation-and-trait-enrichment-workflows)
 - [References](#references)
 
 ---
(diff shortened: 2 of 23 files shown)


eberrigan (changes requested):
## Review Summary

Scientifically sound, well-decomposed, and faithful to the architecture review on this PR — I verified all eight required items from that review (A–H below) are **genuinely implemented**, not just claimed. FDR pooling, the exact-binomial enrichment, the sample-PCA→genotype-mean→correlate ordering, the typed `CrossPlatformPCResult`, the separate `bonferroni`-aware PC validator, and the default-off DAG step are all correct. CI is green on all three OSes.

The one thing holding this back from approval: two **edge-case crashes** that contradict the library's "messy phenotyping data → NaN/empty + WARNING, never crash" contract. Both are one-line guards. Everything else is polish.

## Blocking Issues

**1. Zero common genotypes (or empty result) crashes `cross_platform_pc_correlations`** — `pc_correlations/correlate.py:311-313`
`summarize_correlations` sets `n_genotypes = 0` when a pair shares no genotypes, then calls `minimum_detectable_correlation(0, ...)`, which raises `ValueError: n must be a positive integer, got 0`. The per-PC rows are correctly produced as all-NaN (the `n_genotypes >= 2 and std>0` guard at `correlate.py:143` works), so only this summary line crashes. Fix: when `n_genotypes < 4`, set `min_detectable_r = np.nan` instead of calling the helper (mirrors the NaN path already used upstream).

**2. Empty correlation table hard-crashes the enrichment DAG step** — `pc_correlations/enrichment.py:96-97` via `pipeline/steps/calculate_trait_enrichment.py`
`CalculateCrossPlatformCorrelationsStep` deliberately emits an empty table + WARNING when all trait pairs fall below `min_genotypes_for_correlation` — a *survivable* outcome that the downstream visualize step already handles gracefully (`visualize_cross_platform.py:89`). But with `enrichment_enabled=True`, `count_significant` returns `n_tests=0` → `calculate_enrichment_test` raises `ValueError: n_tests must be positive`, killing the DAG at step 4 before visualize is reached. So enabling the new (documented) feature converts a survivable run into a hard failure. Fix: when `n_tests == 0`, skip the binomial, log a WARNING, and pass `data` through (write an empty/!-skipped CSV) rather than raising.

*(Both are untested — see Important #1. If you judge these inputs genuinely impossible in practice, say so and I'll defer, but the rest of the codebase treats them as survivable.)*

## Important Issues

1. **Missing the representative-population pin test that issue #119 explicitly asks for.** #119's acceptance gate wants the representative population "pinned by a test" so the ~5,027 trait-test count can't silently drift. Good news: I traced item H and the feared "all-rows-with-an-`is_representative`-flag" bug does **not** exist — `ReduceTraitRedundancyStep` filters to representatives upstream and the CSV is structurally representative-only (no flag column), so counting rows directly is correct. But there's no test that pins this end-to-end (the existing test counts rows on a synthetic table; it doesn't assert the clustering path yields the representative count). Add one so a future change to the upstream step can't silently break the count.

2. **Degenerate / edge-case code paths are completely untested.** `correlate.py:143-148` (n<2 / zero-variance → NaN) and `_apply_multipletests`'s NaN-skip (`correlate.py:199-207`) have dedicated branches with zero coverage; likewise the enrichment `n_significant=0` / `n_significant==n_tests` boundaries and the empty-alignment path. These are exactly the branches that hide blockers #1 and #2. Adding empty / single-genotype / all-NaN tests would have caught both.

3. **`count_significant` inflates the binomial denominator with NaN-p rows** — `enrichment.py:186`. `n_tests = len(df)` includes degenerate trait pairs whose p is NaN (correctly excluded from the numerator), biasing the test slightly toward "depleted." Use `n_tests = int(df[p_value_column].notna().sum())`.

4. **`summarize_correlations` hardcodes `alpha=0.05`, `power=0.80`** — `correlate.py:311-313`. Every other call threads the configurable `alpha`, so a caller running `alpha=0.01` gets a headline `min_detectable_r_80_power` computed at the wrong alpha. Thread `alpha` through (and ideally make `power` a named constant).

5. **matplotlib backend is set in `scripts/` but not in the CLI/library path** — `pc_correlations/visualize.py:17`, `enrichment.py:19` import `pyplot` with no backend; the `pc-correlations` CLI command never sets Agg. Headless CI auto-falls-back to Agg (so it's green), but a Windows/macOS GUI dev machine loads the interactive backend → popped windows / init failures. Set `matplotlib.use("Agg")` in the workflow/CLI before the pyplot import.

6. **Figure leak: `fig.clf()` instead of `plt.close(fig)`** — `workflow.py:188,194,197,278`. `clf()` clears contents but leaves the figure registered in pyplot's global manager, so ~9 figures/call (scopes × pairs) are retained for the process lifetime. The rest of the codebase uses `plt.close(fig)` — make this consistent.

7. **All-platform-intersection `n` is undocumented** — `aggregate.py:147-169`. For 3+ platforms, every pair is correlated on the genotypes common to *all* platforms, not the true pairwise intersection, so per-pair `n` (and the headline MDR/power) silently understate available data. Defensible, but document it. Relatedly, with ~19 genotypes the CIs/power are approximate (MDR@80% ≈ 0.58) — a one-line caveat in the workflow docstring would help reproducers.

## Suggestions

- **TDD:** git history shows tests landed in the *same* commit as implementation (`274894c`) and the step tests in the review-refactor commit (`d40b348`) — test-alongside/after, not red-green-refactor. Tests genuinely recompute (not golden-pinned), so quality is good; just flagging the process claim.
- **Doc bug:** `docs/CROSS_PLATFORM_ANALYSIS.md` shows `result["summary"]` for `cross_platform_pc_correlations`, but it returns a frozen `CrossPlatformPCResult` dataclass → use `result.summary` (attribute access). The `trait_correlation_enrichment` example just below correctly uses `["results"][0]` since that one returns a dict.
- **Stale log denominators:** `cross_platform_pipeline.py:167,177,193` still say `Step x/4`; the pipeline now has 5 steps (4/5 and 5/5 were updated, the first three weren't).
- `scripts/run_pc_correlations.py:37` `--fdr-scope` lacks `choices=` (the `click` CLI has it); the workflow validates and fails safely, but parse-time `choices` matches the CLI.
- Consider a skip-guarded enrichment golden (combined 5,027 → depleted, fold≈0.82) mirroring the PC golden, so the headline is pinned before #120 fixtures land.

---

eberrigan (commented):
> **Verdict: COMMENT** — re-review of `e8f83570a` ("harden #119 workflows + review fixes")

## Summary

The earlier `CHANGES_REQUESTED` is effectively cleared: **both blocking edge-case crashes are genuinely fixed**, each with a regression test that would have crashed before the fix, and all 7 important items + the suggestions are addressed. CI is green on all three OSes. Nicely done — this is a faithful, complete response to the review.

### Verified resolved
- **Blocker 1** (zero common genotypes crashed `summarize_correlations`): fixed via the `n_genotypes >= 4` → NaN-MDR guard (statistically correct — Fisher-z SE = 1/√(n−3)); reproduced by `test_pc_workflow_zero_common_genotypes_does_not_crash`.
- **Blocker 2** (empty correlation table killed the enrichment DAG step): fixed via the survivable `n_tests == 0` branch (warn + pass `data` through); reproduced by `test_enrichment_step_empty_table_does_not_crash`.
- **Important 3** (`count_significant` excludes NaN-p rows from the denominator), **4** (`alpha` threaded — confirmed at the call site, plus the `MIN_DETECTABLE_POWER` constant), **5** (CLI sets the Agg backend), **6** (`plt.close(fig)` at all four sites), **7** (intersection-`n` + small-panel caveats documented). Doc/`Step x/5`/`--fdr-scope choices` nits all done.
- I also probed the two most likely regression traps and confirmed they are **non-issues**: the `matplotlib.use("Agg")`-after-pyplot-import ordering (matplotlib ≥3.10 switches cleanly), and the `n_tests==0` early-return metadata contract (the visualize step reads `data["correlation_df"]` + upstream metadata, never the enrichment CSV).

## Two new, non-blocking items the fix introduced

1. **Malformed empty enrichment CSV** — `pipeline/steps/calculate_trait_enrichment.py` (the `n_tests == 0` branch): `results_to_dataframe([]).to_csv(...)` writes a **zero-column file with no header** (`pd.read_csv` raises `EmptyDataError` on it). The DAG doesn't break — nothing reads it back in-pipeline — but it violates the artifact schema, so an external consumer or the `cross-platform-summary` skill would choke on it. The test only asserts the file `.exists()`, so it doesn't catch the missing header. Suggest writing an empty frame with the correct columns (e.g. `pd.DataFrame(columns=[f.name for f in fields(EnrichmentResult)])`).

2. **The representative-count pin test is tautological** — `test_enrichment_representative_count_pinned` asserts `enriched.metadata["enrichment_n_tests"] == corr_df["spearman_p"].notna().sum()`, but the step *computes* `enrichment_n_tests` as exactly that sum on the same frame, so the assertion holds for any input. It doesn't pin a concrete count, so upstream clustering drift (e.g. 19 → 30 representatives) would move both sides together and pass — the only real guard is the loose `exp1_reduced_traits < 50`. This was the original Important #1 ("pin so the count can't drift"); pinning an exact expected integer would actually meet that intent.

Neither blocks merge or the never-crash-on-messy-data contract — both are quick follow-ups.

---

eberrigan (approved):

Both follow-up items from the prior re-review are genuinely resolved (verified at runtime):

- **Schema-conformant empty enrichment CSV** — `results_to_dataframe([])` now returns `pd.DataFrame(columns=[f.name for f in fields(EnrichmentResult)])`, so the `n_tests==0` branch writes a header-bearing CSV (no more `EmptyDataError`). Columns derive from the dataclass fields — same source as the normal path, so they can't drift. The test now reads the CSV back and asserts the header, not just `.exists()`.
- **Concrete count pin** — `test_enrichment_representative_count_pinned` now pins exact integers (`exp1_reduced_traits == 1`, `exp2_reduced_traits == 3`, `enrichment_n_tests == 3`); a reviewer confirmed mutating the pin makes it fail, so it genuinely catches upstream clustering drift.

Combined with the earlier round (both edge-case blockers fixed with reproducing regression tests, all 7 important items addressed), this is ready. CI green on all three OSes. Nice work.

---

PR #149, opened 2026-06-12
feat: add PCAResult serializable dataclass + adapter (#127)

egao28:
## Summary

Adds a stdlib `PCAResult` dataclass (with nested `FeatureContribution`) and a `from_pca_dict()` adapter in a new shared `result_types.py` module. This is the **detailed exemplar** for the serializable-result-types epic (#130); #128 (`HeritabilityResult`) and #129 (`ClusterResult`) will follow the same convention in this module.

`perform_pca_analysis()` returns a dict mixing numpy arrays, pandas frames, and sklearn objects — it can't cross a JSON boundary (bloom-mcp, caching, an API) without bespoke conversion. `PCAResult` is a clean, JSON-native view: `json.dumps(dataclasses.asdict(result))` round-trips with no custom serializer.

Closes #127.

## What's in the result

- Science only — **no sklearn `PCA`/`StandardScaler`** in the view.
- `scores` mapped from the dict's `transformed_data`; every scalar cast to native `int`/`float` (numpy scalars like `np.int64` are **not** reliably JSON-serializable).
- `feature_contributions` preserve the source DataFrame's `total_contribution`-descending order and name-from-index correspondence.
- Carries per-component `cumulative_variance_ratio`; `@property cumulative_variance` for the scalar total.
- `from_pca_dict(d, *, random_state=None, explained_variance_threshold=None)` stamps reproducibility provenance (these aren't in the dict).
- `frozen=True` + `to_dict()`, matching the `EnrichmentResult`/`CrossPlatformPCResult` precedent.

## Backwards-compat

**Additive only (MINOR).** `perform_pca_analysis` still returns its exact dict; the adapter does not mutate its input. `PCAResult`/`FeatureContribution` exported in `__all__`.

## Testing

- New `tests/test_pca_result.py` (13 tests): native-type JSON round-trip (the reproducibility CI gate), no-sklearn-object guard, adapter shape mapping, `feature_contributions` ordering/name fidelity, both `standardize` branches, `n_components=1` nested-shape preservation, provenance stamping, determinism (epic #118), public export, and a dict-unchanged/non-mutating guard.
- Full suite green (2087 passed); `black`/`ruff` clean; `openspec validate add-pcaresult-dataclass --strict` passes.

- `docs/result-types.md` pattern doc (written once at epic close, across all three types).
- Wheat-EDPIE golden numeric assertions (epic verification milestone #120/#130).

The OpenSpec proposal was honed via a 5-agent adversarial review before implementation.

### DIFF
--- a/openspec/changes/add-pcaresult-dataclass/proposal.md
+++ b/openspec/changes/add-pcaresult-dataclass/proposal.md
@@ -1,0 +1,88 @@
+# Add PCAResult Serializable Dataclass Return Type (issue #127)
+
+## Why
+
+`perform_pca_analysis()` (`pca.py`) returns a `Dict` that mixes four
+serializability classes in one bag: clean scalars/lists
+(`n_components_selected`, `feature_names`), `np.ndarray` fields
+(`loadings`, `transformed_data`, `explained_variance_ratio`, `eigenvalues`),
+`pd.DataFrame` fields (`feature_contributions`, `feature_metrics_df`), and
+non-serializable sklearn objects (`pca`, `scaler`). This makes the function
+impossible to expose over any JSON boundary — bloom-mcp (Metcalf 2026), an API,
+or a cached artifact — without bespoke conversion.
+
+The package already has JSON-safe stdlib `@dataclass` result types in
+`summary/cross_platform_summary.py` (`TraitReductionStats`, `PowerStats`,
+`CorrelationStats`, `TopCorrelation`, …) and the `EnrichmentResult` pattern
+(flat, JSON-native, `to_dict()`). This change extends that pattern to PCA as the
+**detailed exemplar** for the serializable-result-types epic (#130), with #128
+(`HeritabilityResult`) and #129 (`ClusterResult`) following the identical
+convention.
+
+## What Changes
+
+- **New module `src/sleap_roots_analyze/result_types.py`** — the shared home for
+  the epic's serializable result types (PCA now; heritability/cluster later). It
+  imports nothing from `pca.py`, keeping dependencies one-way.
+- **`PCAResult` and `FeatureContribution` stdlib dataclasses** exposing **only
+  serializable science** — sklearn `pca`/`scaler` objects stay out of the clean
+  view (still available via the legacy dict for in-process callers).
+  `frozen=True` and a `to_dict()` convenience (`asdict(self)`), matching the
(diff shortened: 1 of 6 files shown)


eberrigan (commented):
> **Verdict: COMMENT**

## Review Summary

Full subagent-team review (Code Quality · Testing · Statistical Rigor · Performance/Memory · Behavioural Correctness) across the #149→#150→#151 stack. No blocking issues. This is the strongest of the three: additive-only (source `pca.py` untouched — verified), correct casting, real `perform_pca_analysis` output in tests, genuine `n_components=1` nested-shape + provenance + determinism tests. `feature_contributions` ordering and name fidelity are correct and well-tested — `iterrows()` binds each name to its values, so `asdict` reordering / `total_contribution` ties cannot mismatch a name with the wrong value (the classic bug is avoided).

## PR-specific (#149)

- The cross-cutting NaN/Inf gap is most reachable here via `eigenvalues`/`scores`/`explained_variance_ratio` pass-through, plus a degenerate all-zero-loadings case where `pca.py` computes `total / np.sum(total)` → NaN `fractional_contribution`.
- Per-feature variance breakdowns (`explained_variance_ratio_per_feature`, `feature_variances`) are dropped from the view; only the sklearn-object omission is documented — add a one-line note that these secondary fields are intentionally dropped.
- Degenerate fixtures (`pca_all_nan_data`, `pca_single_sample_data`, `pca_zero_variance_all_columns`) exist but are unused by this suite.

---

eberrigan (approved):

All prior-review items resolved (verified at runtime, 19 tests pass, CI green on 3 OSes):

- **Float native-type tests are no longer vacuous** — `test_fields_are_native_types_pre_serialization` asserts `type(x) is float` on the **pre-serialization** dataclass fields (not after a JSON round-trip that laundered `np.float64`). Confirmed it would fail if an adapter dropped its `float()`/`.tolist()` cast.
- **NaN/Inf JSON contract enforced** — new `to_json()` sets `allow_nan=False`; parametrized tests assert `ValueError` on nan/±inf, finite round-trip passes.
- **Non-mutation test deep-compares values** (array/frame equality + `id()` identity on sklearn objects), not just `set(d.keys())`.
- Per-feature variance drop + caller-provenance caveat + shallow-`frozen` now documented.

This is the base of the dataclass stack — approving for merge first (#149 → #150 → #151).

---

PR #150, opened 2026-06-12
feat: add HeritabilityResult serializable dataclass + adapter (#128)

egao28:
## Summary

Adds a frozen `HeritabilityResult` dataclass (with nested `TraitHeritability`) and a `from_heritability_dict(d, threshold)` adapter to the shared `result_types.py` module. This is the **Phase-1 anchor** of the serializable-result-types epic (#130), following the `PCAResult` (#127) convention.

`calculate_heritability_estimates()` returns a trait-keyed dict mixing per-trait estimates with a `__calculation_metadata__` entry — awkward to serialize or expose. `HeritabilityResult` is a clean JSON-native view: `json.dumps(dataclasses.asdict(result))` round-trips with no custom serializer.

Closes #128.

> **Stacked on #127** (`pcaresult-dataclass-127`, PR #149), which introduces `result_types.py`. Base will retarget to `main` once #127 merges.

## What's in the result

- Per-trait H², variance components (`var_genetic`/`var_residual`), and counts — all native-cast to JSON-safe types.
- `method` read from `__calculation_metadata__` (that entry is **not** treated as a trait); traits that errored go to `failed_traits`, not `per_trait`.
- `@property mean_h2` (mean over successful traits, `0.0` when empty) and `n_above_threshold`.
- `frozen=True` + `to_dict()`, matching the `PCAResult`/`EnrichmentResult` precedent.

## Backwards-compat

**Additive only (MINOR).** `calculate_heritability_estimates` still returns its exact dict/tuple; the adapter does not mutate its input. `HeritabilityResult`/`TraitHeritability` exported in `__all__`.

## Testing

- New `tests/test_heritability_result.py` (10 tests) over a real `calculate_heritability_estimates` run on the `heritability_data_known_h2` fixture (known variance components): native-type JSON round-trip, method/metadata handling, threshold classification + H² ordering, `mean_h2`/`n_above_threshold`, `failed_traits` separation, empty-case, determinism, export, and a dict-unchanged/non-mutating guard.
- `black`/`ruff` clean; `openspec validate add-heritabilityresult-dataclass --strict` passes; statistics + public-API contract suites green.

- Wheat-EDPIE Turface-19 golden numbers (mean H²=0.77; 8 traits ≥0.60) → epic verification milestone (#120/#130), as with #127. Numeric correctness is anchored here on the synthetic known-H² fixture.
- `docs/result-types.md` pattern doc → epic close.

### DIFF
--- a/openspec/changes/add-heritabilityresult-dataclass/proposal.md
+++ b/openspec/changes/add-heritabilityresult-dataclass/proposal.md
@@ -1,0 +1,65 @@
+# Add HeritabilityResult Serializable Dataclass Return Type (issue #128)
+
+## Why
+
+`calculate_heritability_estimates()` (`statistics.py`, exposed publicly via
+#116) returns a `Dict` keyed by trait name, each value a per-trait dict of
+broad-sense heritability plus variance components and a special
+`__calculation_metadata__` entry. Like the PCA dict, this nested, metadata-mixed
+shape needs a clean serializable view to cross a JSON boundary (bloom-mcp,
+caching, an API) and to anchor the wheat EDPIE golden tests.
+
+This is the Phase-1 anchor of the serializable-result-types epic (#130). It
+follows the exact convention established by `PCAResult` (#127): a frozen stdlib
+`@dataclass` holding only JSON-serializable science, a `from_*_dict()` adapter,
+`frozen=True` + `to_dict()`, and native-Python scalar casts.
+
+## What Changes
+
+- **`HeritabilityResult` and `TraitHeritability` frozen dataclasses** added to
+  the shared `result_types.py` module (created in #127). They expose only
+  serializable science — per-trait H², variance components, and counts — never
+  sklearn/statsmodels objects.
+- **`HeritabilityResult.from_heritability_dict(d, threshold)` adapter** mapping
+  the `calculate_heritability_estimates` return dict (the
+  `remove_low_h2=False` form, or the first element of the
+  `remove_low_h2=True` tuple):
+  - `method` is read from `d["__calculation_metadata__"]["method_used_for_all_traits"]`.
+  - The `__calculation_metadata__` key is skipped; trait entries carrying an
+    `"error"` (or lacking a `"heritability"` value) are collected into
+    `failed_traits` rather than `per_trait`.
+  - Each successful trait becomes a `TraitHeritability` with `passed_threshold =
+    h2 >= threshold`; all scalars cast to native `float`/`int`.
+- **`@property mean_h2`** (mean H² over successful traits) and
+  **`@property n_above_threshold`** (count with `passed_threshold`).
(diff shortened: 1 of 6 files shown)


eberrigan (commented):
> **Verdict: COMMENT**

## Review Summary

Full subagent-team review (Code Quality · Testing · Statistical Rigor · Performance/Memory · Behavioural Correctness) across the #149→#150→#151 stack. No blocking issues. Additive-only (source `statistics.py` untouched — verified), tested on real `calculate_heritability_estimates` output over the `heritability_data_known_h2` fixture. `_HERITABILITY_METADATA_KEY` is a named constant (good — set the precedent for the epic), failed-trait separation and the empty-case `mean_h2=0.0` guard are correct, and the `>=` threshold boundary matches the source filter.

## PR-specific (#150)

1. **Run-level `{"error": ...}` failure is misrepresented as a fake trait.** When `calculate_heritability_estimates` short-circuits on missing columns it returns `{"error": "..."}` with no `__calculation_metadata__`; the adapter routes the string `"error"` into `failed_traits=["error"]`, sets `method="unknown"`, and **discards the actual error message**. Detect a lone top-level `error` key and surface it distinctly rather than treating it as a failed trait. (Flagged independently by three reviewers.)
2. **Adapter field mapping is shallow-tested.** `var_genetic`, `var_residual`, `n_genotypes`, `n_observations`, and `model_type` are mapped in `from_heritability_dict` but never value-asserted — only `h2`/`passed_threshold` are checked, so a swapped `var_genetic`↔`var_residual` (or wrong key) would pass every test. Add a "maps all trait fields from a real entry" assertion.
3. **Docstring wording:** `TraitHeritability` repeats "broad-sense" while the source H² is `var_G/(var_G + var_E/mean_n_reps)` — a genotype-mean/repeatability heritability, not the textbook broad-sense `var_G/(var_G+var_E)`. Inherited from the existing codebase (not a regression here), but the new docstring shouldn't entrench the imprecise term.

---

eberrigan (approved):

All prior-review items resolved (verified at runtime, 18 tests pass, CI green on 3 OSes):

- **Run-level `{"error": ...}` no longer becomes a fake trait** — `from_heritability_dict` detects a lone string-valued top-level `error` key and surfaces it on a new `error` field (verified it correctly distinguishes this from a per-trait failure dict, and from a trait literally named "error"). Test pins `failed_traits == []` / `error == msg`.
- **Field mapping now value-asserted** — `test_maps_all_trait_fields_from_real_entry` checks `var_genetic`/`var_residual`/`n_genotypes`/`n_observations`/`model_type` against the real source entry, plus a `var_genetic != var_residual` swap-guard.
- Cross-cutting: pre-serialization `type() is float` assertions, `to_json(allow_nan=False)`, deep-compare non-mutation — all added.
- "broad-sense" docstring corrected to the accurate genotype-mean/repeatability formula.

Stacked on #149 — merge after it.

---

PR #151, opened 2026-06-12
feat: add ClusterResult serializable dataclasses + adapters (#129)

egao28:
## Summary

Adds a frozen base `ClusterResult` plus `KMeansResult` / `GMMResult` subclasses and `from_kmeans_dict` / `from_gmm_dict` adapters to the shared `result_types.py` module. This is the **final child** of the serializable-result-types epic (#130), following the `PCAResult` (#127) / `HeritabilityResult` (#128) convention.

`perform_kmeans_clustering` / `perform_gmm_clustering` return `np.ndarray`-laden dicts that can't cross a JSON boundary. The epic flagged that KMeans and GMM returns differ substantially, so a single flat type doesn't fit — this models it as a **shared base + two subclasses** (the chosen design), with the `algorithm` field as the discriminator a JSON consumer reads.

Closes #129.

> **Stacked on #128** (`heritabilityresult-dataclass-128`, PR #150) → #127 (PR #149). Base retargets to `main` as the stack merges.

## Shape

- **Base `ClusterResult`**: `algorithm` (`"kmeans"`|`"gmm"`), `n_clusters`, `cluster_labels`, `cluster_sizes`, the three quality metrics, `feature_names`, `random_state`.
- **`KMeansResult`** adds `cluster_centers`, `inertia`.
- **`GMMResult`** adds `cluster_centers` (the GMM `means`), `weights`, `bic`, `aic`, `converged`, `n_iter`, `covariance_type`.

## Reproducibility (#118)

Clustering is stochastic, so `random_state` is **threaded into the adapters and stamped** into the result (the legacy dicts don't carry it). A determinism test asserts same seed → identical `cluster_labels` via the typed view.

## Backwards-compat

**Additive only (MINOR).** Both functions still return their exact dicts; the adapters do not mutate input. `ClusterResult`/`KMeansResult`/`GMMResult` exported in `__all__`.

## Testing

- New `tests/test_cluster_result.py` (8 tests) over the real `kmeans_cluster_result` / `gmm_cluster_result` fixtures: native-type JSON round-trips for both subclasses, adapter field mapping (incl. GMM `means`→`cluster_centers`, `n_components`→`n_clusters`), `random_state` stamping, determinism (#118), export, and non-mutation guards.
- `black`/`ruff` clean; `openspec validate add-clusterresult-dataclass --strict` passes; clustering + public-API introspection contract suites green.

- `docs/result-types.md` pattern doc → epic close (#130).
- Per-sample GMM `probabilities` and KMeans distance arrays omitted from the clean view (available via the legacy dict).

---

## Epic-closing additions (#130)

Because this is the final child of the stack, it also lands the two remaining **epic-level acceptance items** for #130:

- **`docs/result-types.md`** — documents the serializable-dataclass pattern once (stdlib dataclass, JSON-native fields, `from_*_dict` adapters, `@property` derivations, additive backwards-compat, and a test-first recipe for new types).
- **`test_viz_pca_typed_view_golden`** (in `tests/test_pipeline_reproduction.py`) — asserts the #120 wheat-EDPIE golden explained variance against the **`PCAResult` typed view** (not just the legacy dict), across all four platforms, and proves `json.dumps(asdict(...))` round-trips.

With these, all three of #130's epic acceptance items are met once the stack merges: (1) three result types exported with round-trip tests, (2) pattern documented once, (3) golden numbers assert against the typed views.

### DIFF
--- a/docs/result-types.md
+++ b/docs/result-types.md
@@ -1,0 +1,82 @@
+# Serializable result types
+
+Most analytical functions in `sleap_roots_analyze` return a `dict` (or tuple) that
+mixes JSON-hostile members — `numpy.ndarray`, sklearn objects (`PCA`,
+`StandardScaler`, `UMAP`), `pandas.DataFrame`. Those results cannot cross a JSON
+boundary without bespoke conversion, which blocks exposing the functions as MCP tools
+(bloom-mcp), caching results as artifacts, and schema-checkable outputs (FAIR:
+interoperable + reusable).
+
+The **result types** in [`result_types.py`](../src/sleap_roots_analyze/result_types.py)
+solve this: flat stdlib `@dataclass` views that hold **only JSON-serializable science**.
+This is the same house pattern already used in
+[`summary/cross_platform_summary.py`](../src/sleap_roots_analyze/summary/cross_platform_summary.py)
+(`TraitReductionStats`, `PowerStats`, …), extended to the analytical functions under the
+serializable-result-types epic (#130).
+
+## The pattern
+
+A result type is a `@dataclass(frozen=True)` that satisfies these rules:
+
+1. **stdlib `dataclasses`, not Pydantic.** The package stays dependency-light and
+   MCP-agnostic; bloom-mcp wraps these plain dicts in its own Pydantic models at its
+   boundary.
+2. **Every field is JSON-native** — Python scalars, `str`, `bool`, and (nested) `list`s
+   of those. Never `numpy` arrays, `pandas` frames, or sklearn objects. The conversion
+   to native types happens once, in the adapter (`np.asarray(...).tolist()`,
+   `float(...)`, `int(...)`).
+3. **Science only.** Fitted sklearn objects (`PCA`, `StandardScaler`) stay out of the
+   clean view — they remain available via the legacy dict for in-process callers.
+4. **Derivations are `@property`, not stored fields**, so the serialized state has no
+   redundant/derivable members (e.g. `PCAResult.cumulative_variance`,
+   `HeritabilityResult.mean_h2`).
+5. **A `from_*_dict()` classmethod adapter** builds the dataclass from the legacy dict
+   and must **not mutate** the input dict.
+
+Because every field is JSON-native, the invariant that anchors the reproducibility CI
+gate holds with no custom serializer:
+
+```python
+import json, dataclasses
+from sleap_roots_analyze import PCAResult
+
+result = PCAResult.from_pca_dict(perform_pca_analysis(df), random_state=42)
+restored = json.loads(json.dumps(dataclasses.asdict(result)))   # round-trips, no encoder
+```
+
+## The types
+
+| Type | Built from | Adapter |
+| --- | --- | --- |
+| `PCAResult` (+ `FeatureContribution`) | `perform_pca_analysis` dict | `PCAResult.from_pca_dict(d, *, random_state=None, explained_variance_threshold=None)` |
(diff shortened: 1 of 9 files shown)


eberrigan (commented):
> **Verdict: COMMENT**

## Review Summary

Full subagent-team review (Code Quality · Testing · Statistical Rigor · Performance/Memory · Behavioural Correctness) across the #149→#150→#151 stack. No blocking issues. Additive-only (source `clustering.py` untouched — verified). The base `ClusterResult` + `KMeansResult`/`GMMResult` subclass design is clean (correct field ordering, no non-default-after-default `TypeError`), the highest-risk field renames (`means`→`cluster_centers`, `n_components`→`n_clusters`) are explicitly value-tested, and the raw `np.bool_` `converged` is correctly cast to native `bool`. Wrong-dict cross-feed fails loudly (`KeyError`) rather than silently. Cross-OS same-seed determinism was assessed **safe** — same-process, integer-label comparison, sklearn (not numba) — so it is **not** a UMAP-style flake risk.

## PR-specific (#151)

1. **GMM covariances dropped silently.** The view keeps `covariance_type` but drops the actual per-component `covariances` (plus `log_likelihoods` and the per-k `bic_scores`/`aic_scores` sweep). A GMM result advertising `covariance_type="full"` with no covariance values isn't a faithful description of the fitted model — the cluster shapes/orientations are gone. The PR documents omitting per-sample probabilities but not this. Retain them (they're JSON-serializable as nested lists) or document the omission + rationale.
2. **GMM has no determinism test** (only KMeans) — GMM's EM is the more init-sensitive case, so it's the one that most needs the same-seed→identical-labels assertion. Relatedly, GMM `bic`/`aic`/`converged`/`n_iter`/`weights` are only *type*-asserted after round-trip, not *value*-asserted, so a `bic`↔`aic` swap would pass.
3. **Attribution:** `test_viz_pca_typed_view_golden` (added to `test_pipeline_reproduction.py`) tests **PCAResult** (#149's type), not any clustering type — it should live in #149, not the cluster PR. A reviewer tracing "what does the cluster PR test" is misled.
4. **Inline discriminators:** `algorithm="kmeans"`/`"gmm"` are hardcoded string literals (and re-spelled in docstrings), regressing from the `_HERITABILITY_METADATA_KEY` constant precedent #150 set — these are exactly the JSON discriminator a consumer branches on, so they belong in module constants. Also `docs/result-types.md` never mentions the `to_dict()` helper the types provide.

---

eberrigan (approved):

All prior-review items resolved (verified at runtime, 41 tests pass, CI green on 3 OSes):

- **GMM covariances retained** — `GMMResult.covariances` now carries the per-component matrices (native nested lists, no numpy leak; `(n_clusters, n_features, n_features)` shape asserted via `assert_allclose`). The intentional omissions (per-sample probabilities, per-k bic/aic sweep) are now explicitly documented.
- **GMM determinism + value asserts** — `test_gmm_same_seed_identical_serialized` re-runs GMM twice and compares full `asdict` (cross-OS-safe, same-process); `bic`/`aic`/`converged`/`n_iter`/`weights` value-asserted with a `bic != aic` swap-guard.
- **Misplaced PCA golden test moved** to `test_pca_result.py` (gone from `test_pipeline_reproduction.py`).
- **Discriminators are module constants** (`ALGORITHM_KMEANS`/`ALGORITHM_GMM`); `docs/result-types.md` documents `to_dict()`/`to_json()`.
- Cross-cutting native-type/NaN-JSON/non-mutation items all addressed.

Final PR of the stack — merge after #149 → #150.

---

PR #146, opened 2026-06-11
test: full-pipeline reproduction fixtures — all 4 EDPIE platforms (#120)

egao28:
## Summary

Implements the **reproduction harness** for the full pipeline (**QC → viz → cross-platform**) on the real wheat-EDPIE experiment — issue [#120](https://github.com/talmolab/sleap-roots-analyze/issues/120) — for **all four EDPIE platforms**: `turface_19`, `turface_150`, `cylinder`, and `root_core`. One shared golden source so reproductions, downstream tool tests, and generated tests all exercise every step and the numbers never silently drift.

**This PR imports no `sleap-roots-contracts` and merges on its own.** Analysis-input contract conformance is a **follow-up PR** (gated on the `sleap-roots-contracts[pandas]>=0.1.0a1` release).

## What's here

- **`tests/fixtures/` tree**: `README.md` (layout + tolerance + regenerate policy), `harness/` (EDPIE `run_manifest.yaml` + qc/viz/cross_platform configs — the runnable recipe), `real/wheat_edpie/{inputs,expected}` with curated golden for **all 4 platforms**.
- **Curation (~6 MB total)**: commits only artifacts a test asserts against. Excludes `code_snapshot.tar.gz`, logs, per-row `*_data_*.csv` + `cross_platform_exp*_loaded.csv`, and the **oversized per-stage `pipeline_summary.json`** (52 MB cylinder viz / 13 MB turface_150). The assertable viz metrics are extracted into a compact **`viz_pca_metadata.json`** + **`viz_umap_embedding.csv`** (where the viz run produced one — `root_core` disabled UMAP).
- **Tests** (`tests/test_pipeline_reproduction.py`) over `scope="session"` loaders keyed by platform, **parametrized over all 4 platforms**:
  - **QC** — final-data sample count + roles, removed outlier/trait/sample counts, heritability filter (threshold + retained/removed).
  - **viz** — genuine per-stage **PCA re-run**: `perform_pca_analysis` reproduces the golden explained variance via the deterministic eigenvalue spectrum (`rtol=1e-6`); ANOVA/heritability summary with cross-stage H² consistency; UMAP shape/finiteness where present.
  - **cross-platform** — correlations structure + alignment over the **4 manifest pairings** incl. `root_core_vs_cylinder`.
  - **harness configs** — all 8 (4 qc + 4 viz) validate.
- Cross-platform expected is the **single source shared with #119**.

## Per-platform golden (committed values)

[...]

PR #152, opened 2026-06-12
test: analysis-input contract conformance for the EDPIE fixtures (#147)

egao28:
## Summary

Implements #147 — analysis-input **contract conformance tests** for the EDPIE fixtures, the split-out slice from #146 (gated on the `sleap-roots-contracts 0.1.0a1` release). **Tests + dev dependency only; no `src/` changes.**

Adds `tests/test_contract_conformance.py`, which proves the committed post-QC fixtures and the package's canonical examples conform to `validate_analysis_input`. Canonicalization always runs on a **copy** — the frame that feeds QC/viz/cross-platform is never mutated. This previews the #144 load-boundary transform; it does **not** implement the runtime wiring (the `validate_input` flag / `run-all` integration remain #144).

## What it tests

- **Real-data conformance** (parametrized over all 4 platforms `turface_19`, `turface_150`, `cylinder`, `root_core`): rename native roles (`Genotype`→`genotype`, `Barcode`→`sample_id`, `Replicate`→`replicate`) → drop non-trait metadata via `get_trait_columns` (with explicit role kwargs) → `canonicalize_role_dtypes` → `validate_analysis_input(...).raise_for_status()`.
- **Non-mutation guard**: `pd.testing.assert_frame_equal` proves the shared session fixture is untouched.
- **Canonical examples**: every name from `analysis_input_example_names()` (registry-driven, currently `cylinder`, `cylinder_no_replicate`, `field`, `turface`, `genotype_means`) validates as-is.
- **Negative control**: a frame missing the `genotype` role makes `raise_for_status()` raise `ValueError` (`.ok is False`) — proves the green asserts are non-vacuous.
- **Stale-JSON guard**: asserts no `*_validation.json` expected files remain (removed in `73583f9`).

A pre-PR self-review caught one tautological assertion (the intended `replicate`-leak guard was vacuous); replaced with `check.columns.is_unique`, which provably fires under the default-kwargs regression (the numeric `replicate` duplicates).

## ⚠️ Packaging note — please review `uv.lock`

`sleap-roots-contracts 0.1.0a1` is an alpha. Adding it with `uv add --dev --prerelease=allow` resolved the alpha **but also needlessly bumped `pydantic` to `2.14.0a1`** (another prerelease) for the whole resolution. Since contracts only requires `pydantic>=2.7`, I re-pinned pydantic to stable `2.13.4` via `uv lock --upgrade-package pydantic --upgrade-package pydantic-core`. **Net: the lockfile contains exactly one prerelease (`sleap-roots-contracts`); pydantic stays stable.** Worth a glance on the `uv.lock` diff.

## Test results

- New module: **11/11 pass**
- Reproduction goldens (#146/#120) stay green — the proof `run-all` output is unchanged
- Full suite: **2106 passed**
- `black` + `ruff` clean; `openspec validate add-contract-conformance-tests --strict` passes

## OpenSpec

Change `add-contract-conformance-tests` (new `contract-conformance` capability) — all tasks checked off.

Closes #147

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -22,6 +22,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `enrichment_p_value_column` validated against `correlation_method`. When
   enabled it runs a per-pair exact binomial test (nominal p, no FDR) and writes
   `trait_enrichment.csv`; existing runs are unchanged.
+- Analysis-input contract conformance tests (`tests/test_contract_conformance.py`):
+  validate a canonicalized **copy** of each of the four EDPIE post-QC fixtures and every
+  packaged `sleap_roots_contracts` canonical example against `validate_analysis_input`,
+  with a non-mutation guard and a negative control. Adds
+  `sleap-roots-contracts[pandas]>=0.1.0a1` as a dev dependency. No `src/` changes;
+  the #146/#120 reproduction goldens stay green (#147).
 - Expose the eight `statistics.py` functions through the top-level
   `sleap_roots_analyze` namespace and `__all__`, so they can be imported directly
   (e.g. `from sleap_roots_analyze import calculate_heritability_estimates`) instead
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -25,9 +25,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Analysis-input contract conformance tests (`tests/test_contract_conformance.py`):
   validate a canonicalized **copy** of each of the four EDPIE post-QC fixtures and every
   packaged `sleap_roots_contracts` canonical example against `validate_analysis_input`,
-  with a non-mutation guard and a negative control. Adds
-  `sleap-roots-contracts[pandas]>=0.1.0a1` as a dev dependency. No `src/` changes;
-  the #146/#120 reproduction goldens stay green (#147).
+  under both lenient and `strict=True` modes, with a purity guard and a reason-checked
+  negative control. Adds `sleap-roots-contracts[pandas]>=0.1.0a1` as a dev dependency.
+  No `src/` changes; the #146/#120 reproduction goldens stay green (#147).
 - Expose the eight `statistics.py` functions through the top-level
   `sleap_roots_analyze` namespace and `__all__`, so they can be imported directly
   (e.g. `from sleap_roots_analyze import calculate_heritability_estimates`) instead
--- a/openspec/changes/add-contract-conformance-tests/proposal.md
+++ b/openspec/changes/add-contract-conformance-tests/proposal.md
@@ -1,0 +1,54 @@
+## Why
+
+`sleap-roots-contracts 0.1.0a1` is released, so we can now prove the committed EDPIE
+post-QC fixtures and the package's canonical examples conform to the analysis-input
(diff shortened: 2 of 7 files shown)


eberrigan (commented):
> **Verdict: COMMENT**

## Review Summary

Full subagent-team review (Code Quality · Testing · Statistical Rigor · Performance/Memory · Behavioural Correctness). No blocking issues — this is a tight, well-documented test-only + dev-dependency change, and the anti-vacuity design genuinely holds: a reviewer empirically confirmed that using *default* `get_trait_columns` kwargs leaks the numeric `replicate` into the trait set → duplicate columns → `check.columns.is_unique` fails, so the conformance test cannot pass if canonicalization is a no-op. Dependency hygiene is clean (only `sleap-roots-contracts` resolves as a prerelease; pydantic/pydantic-core stay on stable `2.13.4`/`2.46.4`). All four platforms are parametrized and the example check is registry-driven with an empty-registry guard.

## Suggestions (optional, non-blocking)

1. **Strict mode is untested and free to add.** All four post-QC fixtures *and* all five canonical examples also pass under `validate_analysis_input(..., strict=True)`. Parametrizing `strict` in `test_post_qc_fixture_conforms` / `test_canonical_examples_conform` would lock in a meaningfully stronger contract — as written, a regression that only breaks strict passes silently.
2. **Conformance ≠ trait-set correctness.** For `root_core`/`turface_150`, numeric metadata (`Cid, Sid, GID, Ent, Sub, Entry, …`) flows into the validated "trait" set; the contract is permissive about trait names so it passes, but the `trait_cols` non-empty assert can be satisfied by junk columns. Worth a one-line comment so a future reader doesn't over-trust the check.
3. **`test_canonicalization_does_not_mutate_fixture` overclaims.** It deep-copies the session frame *before* calling the build, so it proves `_build_analysis_input` is pure — not that the shared session fixture is protected. The real protection is every caller passing `.copy()` (a convention, not enforced by this test). Either align the docstring with what it proves, or have one test pass the raw session frame and assert it's unchanged.
4. **Transform duplication with #153.** `_build_analysis_input` re-implements in the test the same canonicalization #153 puts in `src` (`validation/input_contract.py`). Once #153 merges, this test should call the production transform so the two recipes can't drift silently.
5. **Spec-wording divergence:** the spec/tasks call for an explicit `"replicate" not in trait_cols` assert; the test relies on `is_unique` instead (stronger in practice, but a literal assert would be more self-documenting). Also consider asserting the negative-control failure *reason* (message contains `genotype`) so any unrelated future validation change can't satisfy it incidentally.

(Now that #146 is merged to main, a rebase will clean up this branch's diff — functionally fine either way.)

---

eberrigan (approved):

All prior-review suggestions addressed (verified at runtime, 16 tests pass, CI green on 3 OSes):

- **Strict mode now tested** — `test_post_qc_fixture_conforms` is parametrized over `strict=[False, True]`; confirmed strict is genuinely stronger (dropping `sample_id` passes lenient, fails strict), so it's a real extra check for all 4 fixtures.
- **Conformance ≠ trait-set correctness** clarified in `_build_analysis_input`'s docstring (names the leaking numeric metadata).
- **Non-mutation test** now passes the **raw** session frame (no pre-copy), so it genuinely proves the shared fixture is unmutated.
- Explicit `assert "replicate" not in trait_cols` (matches spec wording) + negative control now matches `ValueError, match="genotype"`.
- Transform-duplication with #153 correctly **deferred** with a forward-pointer (the production transform doesn't exist until #153 merges).

Ready to merge.

---

PR #153, opened 2026-06-12
feat: wire input-contract validation into QC load boundary (#144)

egao28:
## Summary

Wires the optional `sleap-roots-contracts` `validate_analysis_input` into the **QC data-load boundary** (`LoadDataStep`) so malformed analysis input fails with a clear, structured error **before** any analysis runs. Implements GitHub issue #144 (the *runtime wiring*; fixture conformance is the separate #147).

Validation is a **non-intrusive, optional side-check**: it canonicalizes a **discarded copy** of the entry frame and never touches the data fed to the pipeline, so enabling it never moves a result. analyze still runs when `sleap-roots-contracts` is not installed (logged no-op, never `ImportError`).

## What's included

- **Optional dependency** — `sleap-roots-contracts[pandas]>=0.1.0a1` as a `[contracts]` extra **and** a dev-group member; guarded import (`CONTRACTS_AVAILABLE`), `uv.lock` committed (build.yml uses `--frozen`).
- **Config flag** — `data.validate_input: off | warn | strict` (default `warn`), enum-validated in `validate_qc_config`. The `DataConfig.validate_input` docstring is the canonical description of the modes.
- **Boundary helper** — `sleap_roots_analyze/validation/input_contract.py`: rename config roles → canonical, drop non-trait metadata via `get_trait_columns`, `canonicalize_role_dtypes`, then `validate_analysis_input`. `off` = no-op; `warn` = log warnings + raise only on structural errors; `strict` = raise on any violation.
- **Wiring** — `LoadDataStep.execute` calls the helper once on the entry frame (both CSV and pre-loaded root-core branches); entry-input only, no internal intermediates.
- **Docs** — CHANGELOG `[Unreleased]`, README `[contracts]` install extra, commented `validate_input` in both golden QC templates.

## Scope

Targets the **QC entry boundary only**. The cross-platform boundary is **deferred** to a follow-up: `CrossPlatformConfig` has no barcode/replicate roles and feeds an already genotype-indexed frame, so it needs a distinct genotype-only recipe and its own `validate_cross_platform_config` (OpenSpec task 7.1).

## Acceptance criteria

- [x] `sleap-roots-contracts` is **optional**; run-all runs cleanly when absent (logged no-op, identical output)
- [x] `validate_input: off | warn | strict` flag; default `warn`; `strict` available at the external boundary
- [x] Canonicalize (rename → drop metadata → `canonicalize_role_dtypes`) on a **copy**, then validate; entry input only
- [x] Hard-fail only on missing-`genotype` / no-trait / NaN-genotype; warnings logged, not fatal; trusted golden runs unchanged
- [x] **Equivalence test:** `off` == `warn` → identical loaded frame on the #120/#146 `turface_19` golden
- [x] **Runs-without-contracts test:** no crash, identical output, logged skip
- [x] **Behavior tests:** good passes; malformed warns (`warn`) / raises (`strict`) — packaged examples + malformed fixtures

## Testing

- Full suite: **2116 passed**
- New module: **100% statement + branch coverage** (measured via the coverage API to sidestep the documented scipy/numpy-under-coverage incompatibility on this env)
- black + ruff (pydocstyle/google) + `openspec validate --strict`: all clean

## Relationship

Downstream of contracts#3 (`v0.1.0a1`) · runtime counterpart to fixture-conformance **#147** · cross-platform wiring deferred to a follow-up · validated by #120/#146 golden.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,18 @@ Or with [uv](https://docs.astral.sh/uv/):
 uv add sleap-roots-analyze
 ```
 
+### Optional: input-contract validation
+
+To validate analysis input against the [`sleap-roots-contracts`](https://github.com/talmolab/sleap-roots-contracts)
+schema at the data-load boundary, install the optional `contracts` extra:
+
+```bash
+pip install "sleap-roots-analyze[contracts]"
+```
+
+This enables the `data.validate_input: off | warn | strict` config flag. When the
+extra is not installed, validation degrades to a logged no-op (never an error).
+
 For development:
 
 ```bash
--- a/configs/templates/qc_template_grouped.yaml
+++ b/configs/templates/qc_template_grouped.yaml
@@ -31,6 +31,14 @@ data:
   # Additional columns to exclude from analysis (besides standard metadata columns)
   additional_exclude_cols: []
 
+  # Optional input-contract validation at the data-load boundary (issue #144).
+  # Requires the optional `sleap-roots-contracts` dependency; a logged no-op if it
+  # isn't installed. See the DataConfig.validate_input docstring for details.
+  #   off    = skip validation
+  #   warn   = log issues; hard-fail only on structural errors (default)
+  #   strict = raise on any contract violation (use at untrusted/external boundaries)
+  validate_input: warn
+
   # CRITICAL: Column to group by. Each unique value becomes an independent analysis.
   # Common use: "plant_age_days" for timepoints, "site" for multi-location experiments.
   # Each group must have ≥ min_samples_per_trait samples or it will be skipped.
--- a/configs/templates/qc_template_ungrouped.yaml
+++ b/configs/templates/qc_template_ungrouped.yaml
@@ -31,6 +31,14 @@ data:
   # E.g., ["File.me", "region", "set"] for columns that are not traits but also not standard metadata
   additional_exclude_cols: []
 
+  # Optional input-contract validation at the data-load boundary (issue #144).
+  # Requires the optional `sleap-roots-contracts` dependency; a logged no-op if it
+  # isn't installed. See the DataConfig.validate_input docstring for details.
+  #   off    = skip validation
+  #   warn   = log issues; hard-fail only on structural errors (default)
+  #   strict = raise on any contract violation (use at untrusted/external boundaries)
+  validate_input: warn
+
   # Group-by column for multi-group analysis (set to null for ungrouped analysis)
   group_by: null
 
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -28,6 +28,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   under both lenient and `strict=True` modes, with a purity guard and a reason-checked
   negative control. Adds `sleap-roots-contracts[pandas]>=0.1.0a1` as a dev dependency.
   No `src/` changes; the #146/#120 reproduction goldens stay green (#147).
+- Optional input-contract validation at the QC data-load boundary
+  (`LoadDataStep`). A new `data.validate_input: off | warn | strict` flag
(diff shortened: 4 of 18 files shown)


eberrigan (commented):
> **Verdict: COMMENT**

## Review Summary

Full subagent-team review (Code Quality · Testing · Statistical Rigor · Performance/Memory · Behavioural Correctness). No blocking issues. The core design — canonicalize a discarded copy, validate it, never touch the frame fed to the pipeline — is sound and well-proven: a reviewer vendored the actual `sleap-roots-contracts` 0.1.0a1 source and confirmed `canonicalize_role_dtypes` and `validate_analysis_input` are non-mutating, another ran 329+ existing tests under the new default (`validate_input="warn"`) with contracts installed — all green. The off-vs-warn equivalence test on the real turface_19 golden is exactly the right reproducibility proof. Optional-dependency degradation is covered by a genuine `ImportError`-path test.

## Important (worth addressing, none block merge)

1. **Undocumented behavior change — NaN genotype.** Under the default `warn`, a NaN in the genotype role now hard-fails at the load boundary. This is correct hardening (you don't want stats grouped on unlabeled samples), but the `DataConfig.validate_input` docstring lists the structural hard-fails as "missing genotype / no numeric trait / bad role dtype" and **omits NaN-genotype**. Add it to the docstring + a CHANGELOG line, since real field/Bloom exports do carry blank genotype cells and those configs will now error by default.
2. **Invalid `mode` silently behaves as `warn`.** `validate_entry_input` only branches on `== "off"` / `== "strict"`, so a programmatic caller passing `mode="foo"` (bypassing `validate_qc_config`) gets warn semantics rather than an error. The helper is public with a documented three-value contract — a top-of-function `if mode not in {"off","warn","strict"}: raise ValueError(...)` closes the gap (currently untested).
3. **Unactionable error messages** for two misconfig cases: (a) a wrong `columns.genotype` name produces "required column 'genotype' is missing" instead of naming configured-vs-available columns (contrast `load_trait_data`'s helpful message); (b) a rename collision (df already carries a canonical-named column) yields a bare "duplicate column names" error. Both fail safe — only the messages are poor.

## Suggestions (optional)

- Magic strings (`{off,warn,strict}` + canonical role names) are triplicated across `input_contract.py` / `utils.py` / `components.py` — a shared constant would remove drift risk (and avoid a 4th copy if you add the guard in #2).
- `columns: Any` loses type safety — a `Protocol` or `TYPE_CHECKING` import of `ColumnConfig` documents intent without a runtime import cost.
- The `importlib.reload` contracts-absent test is functionally clean but leaves `load_data` holding the pre-reload function object; it's fragile under `pytest-xdist`. Monkeypatching `CONTRACTS_AVAILABLE` (as the LoadDataStep test already does) would be safer.
- Consider extending the equivalence assertion to `strict` on clean input, so "mode never changes `r.data`" is airtight across all three modes.

---

eberrigan (approved):

All prior-review items resolved (verified at runtime, CI green on 3 OSes):

- **NaN-genotype hard-fail documented** in the `DataConfig.validate_input` docstring + CHANGELOG (with the field-export caveat + remediation).
- **Invalid `mode` now raises** `ValueError` (guard before the `off`/no-op branches, so the no-op is preserved); test parametrized over `["", "Warn", "foo", "none", "off ", None]`.
- **Actionable errors** for both the misconfigured-genotype-name case (names configured vs available columns) and the rename-collision case (clear conflict message) — both fire ahead of any bare contract error; tested.
- **Shared constants** `VALIDATE_INPUT_MODES` / `CANONICAL_ROLES` (single source, re-exported, consumed in `utils.py`).
- `columns: Any` → a `ColumnRoles` Protocol; equivalence proof extended across off/warn/strict; the contracts-absent test moved off `importlib.reload` to monkeypatch (xdist-safe).

This is the base of the contract chain — merge first (#153 → #155).

---

PR #155, opened 2026-06-12
feat: wire input-contract validation into cross-platform loader (#154)

egao28:
## Summary

Completes issue #154 — the cross-platform follow-up to #144/#153. Wires the **same** optional `sleap-roots-contracts` validation into the cross-platform load boundary (`LoadCrossPlatformDataStep`), which PR #153 deliberately deferred.

**Stacked on #153** (`add-input-contract-validation`) — it reuses #153's `validate_entry_input` helper. Review/merge #153 first; this PR's diff is only the #154 commits on top.

## What's included

- **Config flag** — `validate_input: off | warn | strict` (default `warn`) on `CrossPlatformConfig`, enum-validated in its existing `__post_init__`.
- **Helper** — `validate_cross_platform_experiment`, a thin wrapper over `validate_entry_input` with fixed canonical role names. The aligned experiment frames already carry canonical `genotype`/`replicate` columns, so the recipe is identical: canonicalize-then-validate on a discarded copy.
- **Wiring** — `LoadCrossPlatformDataStep.execute` validates **each** loaded experiment frame (exp1, exp2) once, after load/align, before trait selection. The frames fed downstream are never modified.
- **Docs** — CHANGELOG `[Unreleased]` note.

## Why this shape

`CrossPlatformConfig` has no barcode/replicate roles and `load_and_align_experiments` already standardizes the genotype/replicate columns, so the aligned frames are effectively canonical. That let me reuse #153's helper directly (genotype/replicate canonical role names, no `sample_id`) rather than build a separate recipe. The enum check lives in the existing `__post_init__` — no new config-validator function needed.

## Reproducibility / acceptance

- [x] `validate_input: off | warn | strict` (default `warn`) on `CrossPlatformConfig`; invalid value rejected
- [x] Validates each experiment frame on a **copy**; frames fed to alignment/correlation unchanged
- [x] **Equivalence:** `off` == `warn` → identical exp1/exp2 output
- [x] **Runs-without-contracts:** monkeypatched-absent run → no crash, identical output, logged skip
- [x] **Behavior:** good passes under `warn`; missing-`genotype` raises even under `warn`
- [x] Optional dependency unchanged (logged no-op when absent)

## Testing

- New tests in `tests/test_input_contract_validation.py` (helper) and `tests/test_step_load_cross_platform_data.py` (wiring + equivalence + contracts-absent)
- black + ruff (pydocstyle/google) + `openspec validate --strict`: clean
- Note: a few `test_run_all_cli_group_by.py` integration tests are flaky under full-suite load (subprocess contention) and pass in isolation; unrelated to this change. CI is the authoritative gate.

## Relationship

Follow-up to #144 / PR #153 (QC-only wiring) · same optional-contract pattern applied to the cross-platform loader · closes #154.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,8 +28,10 @@ schema at the data-load boundary, install the optional `contracts` extra:
 pip install "sleap-roots-analyze[contracts]"
 ```
 
-This enables the `data.validate_input: off | warn | strict` config flag. When the
-extra is not installed, validation degrades to a logged no-op (never an error).
+This enables the `data.validate_input: off | warn | strict` config flag for the QC
+pipeline, and the matching `validate_input` flag on `CrossPlatformConfig` for the
+cross-platform pipeline. When the extra is not installed, validation degrades to a
+logged no-op (never an error).
 
 For development:
 
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -40,6 +40,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   hard-fail (alongside missing `genotype`, no numeric trait, and bad role dtype), so
   exports carrying blank genotype cells will error unless `validate_input` is set to
   `off` or the column is cleaned first. (#144)
+- Extend the same optional input-contract validation to the **cross-platform**
+  load boundary (`LoadCrossPlatformDataStep`). New `validate_input: off | warn |
+  strict` flag (default `warn`) on `CrossPlatformConfig` validates each loaded
+  experiment frame on a discarded copy; no golden output change. Aligned frames carry
+  no per-sample id, so `strict` injects a synthetic positional `sample_id` into the
+  discarded copy (rather than failing on the structurally-absent role) and otherwise
+  enforces the full contract. Rows with a blank/NaN genotype are now dropped during
+  alignment, so `off`/`warn`/`strict` stay output-identical. (#154)
 - Expose the eight `statistics.py` functions through the top-level
   `sleap_roots_analyze` namespace and `__all__`, so they can be imported directly
   (e.g. `from sleap_roots_analyze import calculate_heritability_estimates`) instead
--- a/docs/CROSS_PLATFORM_ANALYSIS.md
+++ b/docs/CROSS_PLATFORM_ANALYSIS.md
@@ -71,7 +71,29 @@ power_analysis_power: 0.80          # Target power for minimum detectable effect
 top_n_correlations: 30
 top_n_joint_plots: 10
 top_n_boxplots: 10
+
+# Input-contract validation (requires the optional `contracts` extra)
+validate_input: "warn"              # Options: "off", "warn", "strict"
 ```
+
+### Input-contract validation (`validate_input`)
+
+When the optional [`sleap-roots-contracts`](https://github.com/talmolab/sleap-roots-contracts)
+extra is installed (`pip install "sleap-roots-analyze[contracts]"`), each loaded and
+aligned experiment frame is validated on a **discarded copy** at the load boundary —
+it never alters the data fed to the analysis, so the mode never changes results. When
+the extra is absent, validation degrades to a logged no-op.
+
+- `off` — skip validation entirely.
+- `warn` (default) — log non-fatal issues; raise only on universal structural errors
+  (missing/blank `genotype`, no numeric trait, bad role dtype). Rows with a blank
+  genotype are dropped during alignment (they cannot participate in any
+  genotype-aligned comparison), so they never trigger a failure.
+- `strict` — escalate recommended-column issues to errors. Aligned frames carry no
+  per-sample id (the source barcode is dropped during alignment), so `strict` injects a
+  synthetic positional `sample_id` into the discarded copy rather than failing on the
+  structurally-absent role; it otherwise enforces the full contract. For routine runs,
+  `warn` is recommended.
 
 ---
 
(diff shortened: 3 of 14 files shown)


eberrigan (changes requested):
## Review Summary

Full subagent-team review (Code Quality · Testing · Statistical Rigor · Performance/Memory · Behavioural Correctness). The mechanics are correct and faithful to the #153 pattern — copy-isolation, off-no-op, called-once-per-frame, and both-aligned-frames-unchanged are all verified, and the Stat/Perf lenses found nothing blocking. But the Code-Quality, Testing, and Behavioural lenses each independently surfaced real issues, three of which I'd like resolved before merge. (Note this PR is stacked on #153's branch — it needs #153 to merge first, or a rebase onto main.)

## Blocking Issues

**1. `additional_exclude` is not plumbed through (correctness gap).** `load_cross_platform_data.py:81-82` calls `validate_cross_platform_experiment(exp1_df, mode=...)` with no excludes, while the next lines call `get_trait_columns(..., additional_exclude=config.exp1_exclude_cols/exp2_exclude_cols)` for the real analysis. So validation validates a **different column set** than the pipeline analyzes: inside `_build_validation_frame`, `get_trait_columns` runs without the user's excludes, and any numeric excluded-metadata column gets validated as a trait. The PR's own test data demonstrates this exactly (`exp1_exclude_cols=["Ent","Sub"]` with numeric `Ent`/`Sub`). Result: spurious warnings under `warn` (or spurious failures under `strict`), and a numeric metadata column could even mask a genuine "no numeric trait" error. The helper already accepts `additional_exclude` — pass `config.exp1_exclude_cols` / `config.exp2_exclude_cols` into the two calls. The QC path (#153) does pass it; this is an inconsistency, not a design choice. (Flagged independently by two reviewers.)

**2. The strict-mode asymmetry — the novelty of this PR — is untested, and tasks/code contradict.** Aligned frames never carry `sample_id`, so under `warn` the missing-`sample_id` is a recommended-column warning (passes), but under `strict` it raises on **every legitimate aligned frame**. `tasks.md` 2.1 says "good passes under warn/**strict**" (checked off), while the helper docstring says strict *escalates the absent sample_id to an error*. Those contradict — a sign the strict path was never run. No test exercises warn-vs-strict on the sample_id-less frame (compare #153, which tests both directions for the QC helper). Add `test_cross_platform_*_under_strict` and reconcile the contradiction.

**3. NaN-genotype behavior regression under default `warn`.** Reproduced empirically: when both experiments have a blank genotype cell, `load_and_align_experiments` keeps the resulting NaN-genotype rows in the aligned frame, and the validator hard-fails on NaN genotype even under `warn`. Pre-PR those rows were harmless (`groupby("genotype")` dropped the NaN group); post-PR the default aborts a run that previously produced output, which contradicts the spec's own "off == warn produce identical output" promise. Recommend dropping NaN-genotype rows in alignment before validation, or explicitly documenting + testing the new fatal behavior.

## Important Issues

- **`strict` is a config footgun.** `CrossPlatformConfig.validate_input`'s comment claims "same off|warn|strict semantics as DataConfig," but `strict` is functionally unusable on the cross-platform path (guaranteed failure on valid data, per #2). Either reinterpret strict for aligned frames (don't require `sample_id`), or surface the caveat on the field doc rather than only in the `input_contract.py` docstring.
- **Documentation parity gap.** Unlike #153 (which updated README + both QC templates), #155 adds no cross-platform template field and no README / `CROSS_PLATFORM_ANALYSIS.md` mention of the new flag.

## Suggestions

- Mode-validation is duplicated in two idioms (`CrossPlatformConfig.__post_init__` here vs `validate_qc_config()` from #153), each spelling out `["off","warn","strict"]` and a near-identical message. Extract a shared `VALIDATE_INPUT_MODES` constant + helper.
- `_ALIGNED_CANONICAL_ROLES` as a `SimpleNamespace` is correct but reads as a hack; a real `ColumnConfig(genotype="genotype", barcode="sample_id", replicate="replicate", image_path="image_path")` would be self-documenting and type-checked.
- Add a YAML-load-path test (`load_cross_platform_config` with `validate_input: lenient`) — current config tests only construct the dataclass directly; `OmegaConf.to_object` does invoke `__post_init__`, but nothing pins that.
- Consider a test asserting validation doesn't pre-empt the existing "No common genotypes found" error.

Nice work overall — the copy-isolation and no-op machinery is solid; items 1–3 are the substantive ones.

---

eberrigan (approved):
## Re-review verdict: APPROVE ✅ — all 3 blocking items resolved

Verified at runtime (72 tests across the affected files pass with `sleap-roots-contracts` installed, so strict paths are genuinely exercised; CI green on 3 OSes):

1. **`additional_exclude` plumbed through** — the loader now passes `config.exp1_exclude_cols` / `config.exp2_exclude_cols` to the respective validate calls (correct exp1/exp2 ordering, no cross-wiring); test pins `[["Ent","Sub"], ["scanner"]]`.
2. **Strict-mode asymmetry resolved** — `validate_cross_platform_experiment` injects a synthetic positional `sample_id` on a copy for non-`off` modes, so strict enforces the rest of the contract instead of guaranteed-failing on the structurally-absent role (input frame untouched — tested). The tasks.md/docstring contradiction is reconciled, and a no-trait frame still raises under strict (synthesis doesn't mask real errors). The `SimpleNamespace` stand-in became a typed `_AlignedRoles` dataclass.
3. **NaN-genotype regression fixed** — `load_and_align_experiments` drops NaN-genotype rows *before* validation runs, so the default `warn` no longer aborts a previously-working run. Verified this changes no correlation output (those rows were already excluded by every downstream `groupby`/intersection), and the regression was real without the fix.

Also addressed: shared `VALIDATE_INPUT_MODES` constant, strict footgun documented on the config field + README/`CROSS_PLATFORM_ANALYSIS.md`, and YAML-load-path config coverage.

Stacked on #153 — merge after it. Great turnaround on the blocking items.

---

PR #156, opened 2026-06-15
ci: add reproducibility + serialization gates (#133)

egao28:
Closes #133.

## Summary

Adds a dedicated **`reproducibility-gates`** CI job that makes two FAIR guarantees — on which the wheat EDPIE golden tests and bloom-mcp delegation depend — enforceable on every push and PR:

1. **Determinism** — every stochastic analytical function, run twice with a fixed `random_state`, produces identical output.
2. **Result-object serialization** — every result *dataclass* round-trips through `json.dumps(asdict(...))` with no custom encoder.

Both gates already existed in spirit across the #118 (seed handling) and #130/#127/#128/#129 (result-types epic) work; this PR consolidates them into a shared, self-enforcing registry and wires them into CI as a required-able status check. The change is **additive and test/CI-only** — no `src/` runtime behavior changes.

## What changed

| Area | File | Change |
|---|---|---|
| CI | `.github/workflows/ci.yml` | New `reproducibility-gates` job (single OS, 15-min timeout) running the two gate suites |
| Determinism registry | `tests/reproducibility_cases.py` | Shared `CASES` registry of stochastic functions + comparison helpers (new) |
| Determinism gate | `tests/test_reproducibility.py` | Rewritten around the registry: same-seed double-run sweep + coverage guards |
| Serialization gate | `tests/test_result_serialization.py` | JSON round-trip gate over result objects, opt-in by construction (new) |
| Spec | `openspec/changes/add-ci-reproducibility-gates/*` | Proposal + design + spec + tasks (`--strict` valid) |
| Docs | `docs/reproducibility.md`, `docs/CONTRIBUTING.md`, `docs/CHANGELOG.md` | Document the gates + the serialization contract |

### Determinism sweep
A single shared registry (`reproducibility_cases.CASES`) drives the sweep. Functions covered today:

- `pca.perform_pca_analysis`, `pca.perform_pca_with_variance_threshold`
- `umap.perform_umap_analysis`
- `clustering.perform_kmeans_clustering`, `clustering.perform_gmm_clustering`
- `outlier_detection.detect_outliers_isolation_forest`
- (`clustering.perform_hierarchical_clustering` is asserted deterministic separately — no seed)

For each case `test_same_seed_is_reproducible` runs the function twice with the same seed and asserts identical output, and `test_accepts_random_state_none` confirms the `None` path still runs.

**Self-enforcing coverage.** The registry is cross-checked against an explicit `EXPECTED_QUALNAMES` set (and an `EXCLUDED` set), with a `test_coverage_guard_detects_missing_function` meta-test — so a refactor that silently drops a case fails loudly instead of passing with fewer parametrized runs. A newly added stochastic function must be registered (or explicitly excluded), or the guard fails.

### Result-object serialization gate
`test_result_serialization.py` builds one round-trip case per analytical function (the stochastic registry plus the statistics/heritability functions) and asserts `json.dumps(asdict(result))` round-trips. It is **opt-in by construction**: if a function still returns a plain dict, the case `pytest.skip`s with a clear reason. Listing the statistics functions now (which still return dicts) means the gate *begins enforcing automatically* the moment they adopt a result dataclass — no test edits needed. This is why the suite reports skips: those are functions that haven't adopted the result-object pattern yet (plus UMAP when the optional dep is absent).

## Design decisions / notes for future consideration

- **Single OS for this job.** Determinism here is a *same-machine* double-run comparison and serialization is OS-independent, so the matrix `tests` job already covers cross-OS concerns; running the gates once keeps them fast and lets the job be a clean required check.
- **Own job, not folded into `tests`.** Isolating it means it can be set as a required status check in branch protection and its pass/fail is legible at a glance ("reproducibility broke") rather than buried in the matrix.
- **Opt-in serialization gate over a hard requirement.** Avoids blocking functions that haven't adopted the result-types pattern; it grows organically with #127/#128/#129 and future result types.
- **Registry as the single source of truth** shared by both gates avoids drift between "what we claim is deterministic" and "what we serialize."
- **Not included (the issue's optional item):** the scheduled/`run-integration` job to actually run the 100%-skipped integration tests (#69). Left as a deliberate follow-up to keep this PR scoped to the two core gates.

## API changes

None. This PR adds CI configuration, tests, an OpenSpec change, and docs only — no changes to `src/sleap_roots_analyze` runtime behavior or public API.

## Acceptance (issue #133)

- [x] Determinism sweep test added and enforcing in CI.
- [x] Round-trip test covers all current result objects; auto-extends as new ones land.
- [ ] *(Optional)* integration tests run on a schedule / `run-integration` label — deferred (see above).

## Verification

- `uv run pytest tests/test_reproducibility.py tests/test_result_serialization.py` → **32 passed, 15 skipped** locally.
- `openspec validate add-ci-reproducibility-gates --strict` passes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,7 @@ jobs:
         run: uv run ruff check src/sleap_roots_analyze
 
   reproducibility-gates:
-    name: Reproducibility gates
+    name: Reproducibility gates (determinism)
     runs-on: ubuntu-latest
     timeout-minutes: 15
     steps:
@@ -53,12 +53,40 @@ jobs:
       - name: Install dev dependencies
         run: uv sync --group dev
 
-      # Determinism is a same-machine comparison and serialization is
-      # OS-independent, so a single OS is sufficient. Runs as its own job so it
-      # can be configured as a required status check in branch protection.
-      - name: Run reproducibility + serialization gates
-        run: |
-          uv run pytest tests/test_reproducibility.py tests/test_result_serialization.py
+      # Determinism is a same-machine double-run comparison, so a single OS is the
+      # correct, achievable claim (cross-OS bit-identity is not guaranteed by NumPy
+      # /BLAS). Own job so it can be a required status check in branch protection.
+      - name: Run determinism sweep
+        run: uv run pytest tests/test_reproducibility.py
+
+  serialization-gate:
+    name: Serialization round-trip gate
+    timeout-minutes: 15
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+
+      - name: Set up uv
+        uses: astral-sh/setup-uv@v6
+        with:
+          enable-cache: false
+
+      - name: Set up Python
+        run: uv python install 3.11
+
+      - name: Install dev dependencies
+        run: uv sync --group dev
+
+      # Serialization is NOT OS-independent — Path -> str differs by OS
+      # (WindowsPath uses backslashes), exactly the class of bug this gate exists to
+      # catch — so the round-trip runs on the full OS matrix.
+      - name: Run result-object serialization round-trip
+        run: uv run pytest tests/test_result_serialization.py -rs
 
   tests:
     timeout-minutes: 30
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,6 +34,32 @@ jobs:
       - name: Run Ruff
         run: uv run ruff check src/sleap_roots_analyze
 
+  reproducibility-gates:
+    name: Reproducibility gates
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+
+      - name: Set up uv
+        uses: astral-sh/setup-uv@v6
+        with:
+          enable-cache: false
+
+      - name: Set up Python
+        run: uv python install 3.11
+
+      - name: Install dev dependencies
+        run: uv sync --group dev
+
+      # Determinism is a same-machine comparison and serialization is
+      # OS-independent, so a single OS is sufficient. Runs as its own job so it
+      # can be configured as a required status check in branch protection.
+      - name: Run reproducibility + serialization gates
+        run: |
+          uv run pytest tests/test_reproducibility.py tests/test_result_serialization.py
+
   tests:
     timeout-minutes: 30
     strategy:
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -48,6 +48,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   discarded copy (rather than failing on the structurally-absent role) and otherwise
   enforces the full contract. Rows with a blank/NaN genotype are now dropped during
   alignment, so `off`/`warn`/`strict` stay output-identical. (#154)
+- CI reproducibility gates (#133): a whole-package coverage guard that fails CI if any
+  function accepting `random_state` is missing from the determinism sweep
+  (`tests/reproducibility_cases.py`), and an opt-in result-object JSON round-trip gate
+  (`tests/test_result_serialization.py`) that asserts losslessly when a registered
+  function returns a dataclass (and is guarded so a new registered case can't silently
+  drop). The determinism sweep runs single-OS (same-machine comparison); the
+  serialization round-trip runs on the full OS matrix. See
+  [docs/reproducibility.md](reproducibility.md).
 - Expose the eight `statistics.py` functions through the top-level
   `sleap_roots_analyze` namespace and `__all__`, so they can be imported directly
   (e.g. `from sleap_roots_analyze import calculate_heritability_estimates`) instead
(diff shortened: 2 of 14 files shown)


eberrigan (changes requested):
## Review Summary

Full subagent-team review (Code Quality · Testing · Statistical Rigor · Performance/Memory · Behavioural Correctness). The reproducibility-gate design is strong — the determinism registry with its whole-package coverage guard is the best part — and credit where due: **the new round-trip gate caught a real, pre-existing cross-platform serialization bug.** That's the gate working as intended. But CI is red on that genuine bug and the fix isn't in the PR yet, plus a few gate-scoping items, so requesting changes.

## Blocking

**1. Windows CI fails on a real serializer bug — fix it in `src`, with `as_posix()`, not in the test.** Root cause: `pipeline/summary.py` `to_json()` does `"files_generated": [str(f) for f in step["files_generated"]]` *before* calling `convert_to_json_serializable`. On Windows `str(WindowsPath("out/a.csv"))` → `out\a.csv`, and since the value is now a `str`, the serializer's correct `Path → obj.as_posix()` branch (`data_utils.py`) never fires. The test assertion `== ["out/a.csv"]` is **correct** — don't weaken it. Fix: delete the `str(f)` pre-pass and let `convert_to_json_serializable` normalize Paths:
```python
def to_json(self, indent: int = 2) -> str:
    data = convert_to_json_serializable(self.to_dict())
    return json.dumps(data, indent=indent)
```
Note `summary.py` isn't currently in this PR's file list — this src fix needs to be added here so the gate the PR introduces can go green.

### Scope: minimal fix here, systemic sweep as a follow-up
The same `str(path)` anti-pattern is repeated at ~20 sites across the pipeline (e.g. `calculate_cross_platform_correlations.py:393`, all of `reduce_trait_redundancy.py` / `visualize_cross_platform.py`, and `metadata["..."] = str(path)` in `filter_heritability.py`, `merge_all_traits.py`, `visualize_depth_profiles.py`), because `files_generated: List[str | Path]` is a union that lets each producer choose. **Recommend the minimal `summary.py` serializer fix in THIS PR** (it's the bug the gate caught, and the PR's own test needs it), and the **full sweep + tightening `List[str | Path]` → `List[Path]` as a separate follow-up** — it's a cross-pipeline refactor that deserves its own review, and once this PR's gate runs cross-OS (item 3) it's exactly what verifies that sweep. Confirmed low risk: no committed golden asserts on serialized paths and none carry backslash paths, so normalization won't disturb the #146 fixtures. Tracking issue filed.

## Important

2. **"Auto-extends as result types land" is overstated.** The determinism registry has a real package-walking coverage guard; the serialization gate uses a **hardcoded `RT_CASES`/`_STATS_RT_CASES` list with no membership guard**, so a genuinely new result type in a new function is silently uncovered while CI stays green. Add a parallel coverage guard, or correct the wording. Relatedly, `assert_round_trips` **never calls `from_dict`** despite the spec promising "`from_dict` reconstructs an equal object" — that contract ships dead for real types; assert it when `from_dict` is present.

3. **The serialization gate is single-OS, which blinds it to the bug it just hit.** `reproducibility-gates` runs `ubuntu-latest` only, and the `ci.yml` comment claims "serialization is OS-independent" — this PR disproves that. On Linux the gate goes green on the Windows-broken serializer; only the normal Windows *matrix* job caught it. Run the **serialization** round-trip on the OS matrix (determinism can stay single-OS — same-machine comparison is the correct, achievable claim), and fix the comment.

4. **The determinism coverage guard only sees functions that *expose* `random_state`.** Functions that hardcode their seed are invisible: `calculate_mahalanobis_distances` (robust path → `MinCovDet(random_state=42)`) and `detect_outliers_pca`. There's also a real #118-class gap — `detect_outliers_mahalanobis` doesn't forward `random_state` to `calculate_mahalanobis_distances`, so the robust path's RNG isn't caller-controllable and is never exercised. `EXCLUDED` exists as an escape hatch but is empty and can't even document a hardcoded-seed function today.

## Suggestions

- NaN/Inf still round-trip via default `json.dumps` (allow_nan=True) — passes in-process but breaks strict consumers (same issue as #149–#151). Add an `allow_nan=False` assertion or scope the contract in the docstring.
- Dead idempotency assertion in `assert_round_trips` (re-encodes an already-JSON-native tree — always true); run the gate with `-rs` so the currently-large skip set is visible; dedupe the `_silence` helper; label the two deterministic-by-construction cases (`select_n_components`, non-robust mahalanobis) whose seed isn't load-bearing.
- A couple of reviewers saw unrelated bundled changes when diffing `origin/main...HEAD` (likely a stale-base artifact, since GitHub's file list shows 11 files) — worth rebasing on current `main` so the diff is clean.

---

eberrigan (approved):

All items from the request-changes round are resolved, and CI is now fully green on every job — including `Tests (windows)` and `Tests (ubuntu)`, the full-matrix jobs that originally caught the bug.

- **Serializer path bug (was blocking)** — `summary.py:to_json()` now defers to `convert_to_json_serializable` (which normalizes `Path → as_posix()`); the `str(f)` pre-pass is gone. Serialization gate passes on ubuntu/windows/macOS.
- **Gate scoping** — the serialization round-trip gate now runs as a 3-OS matrix (so it can actually catch cross-platform path bugs); determinism stays single-OS (the correct same-machine claim).
- **Seed threading (#118)** — `calculate_mahalanobis_distances` / `detect_outliers_pca` now expose `random_state` (so the coverage guard discovers them) and `detect_outliers_mahalanobis` forwards it; all backward-compatible (defaults preserved).
- **`from_dict` contract** — `assert_round_trips` now asserts `from_dict` reconstructs an equal object when present; the "auto-extends" claim and the NaN/Inf (`allow_nan=True`) scope are now documented honestly rather than overstated.

The path-serialization sweep across the other ~20 sites is tracked separately in #157 and is guarded by this PR's now-cross-OS gate. Nicely done — closes #133.

---

PR #158, opened 2026-06-16
ci: add mypy with a frozen baseline (block new untyped public code) (#132)

egao28:
## Why

There's no static type checking in CI. Full `mypy --strict` on a numpy/pandas/sklearn codebase would red-wall every PR on day one (**329 current errors**), but type quality directly enables the introspection-readiness work (#117) and result-object typing (#130). This takes the pragmatic middle path from #132: **freeze existing type debt in a committed baseline, fail CI only on new errors, and ratchet strictness later.**

Closes #132.

## What changed

- **`pyproject.toml`** — added `mypy` + `mypy-baseline` to the `dev` group and a lenient `[tool.mypy]` block scoped to `src/sleap_roots_analyze` (`ignore_missing_imports`, with `disallow_untyped_defs` as the single starting ratchet knob).
- **`.mypy-baseline.txt`** — froze the 329 pre-existing errors so they don't block PRs.
- **`.github/workflows/ci.yml`** — new dedicated `type-check` job (ubuntu-only; can be made a required status check), modeled on the existing `reproducibility-gates`/`serialization-gate` jobs.
- **`docs/CONTRIBUTING.md`** — documented the ratchet (local command, frozen-baseline meaning, new-defs-must-be-typed, how to regenerate when debt is paid down) plus a pre-submit checklist entry.

No bespoke tooling: `mypy-baseline` is a standard, maintained PyPI tool, so the gate survives maintainer handoff.

## Acceptance (#132)

- [x] `mypy` runs in CI against a committed baseline.
- [x] New untyped public defs fail; existing debt is frozen, not blocking.
- [x] A note in CONTRIBUTING on how the ratchet works.

## How it was verified

- `filter` exits **0** on a clean tree (`new=0`) — existing debt fully absorbed.
- A throwaway untyped public def made `filter` exit **1** (`new=1`); typing it (or removing it) restores green. Reverted.
- `black --check` and `ruff check` still pass.

## Two non-obvious calls

1. **pipefail.** GitHub Actions runs bash with `pipefail` on, and mypy *always* exits non-zero while the baseline holds errors. The job does `set +o pipefail` so the step's pass/fail comes from `mypy-baseline filter` (non-zero only on *new* errors), not mypy.
2. **Baseline path.** mypy-baseline's default file is `mypy-baseline.txt` (no dot); standardized on the dotfile `.mypy-baseline.txt` via explicit `--baseline-path` in CI and docs so the two never diverge.

## Follow-ups (out of scope)

- Tighten strictness incrementally (`disallow_incomplete_defs`, `check_untyped_defs`, per-library stubs) — each a small PR with a baseline refresh.
- Optionally wire mypy into the `/lint` and `/run-ci-locally` slash commands.
- Maintainer: add `type-check` to required status checks in branch protection.


eberrigan (commented):
> **Verdict: COMMENT**

## Review Summary

Full subagent-team review (Code Quality · CI-Gate Discipline · Reproducibility · Performance/Portability · Behavioural Correctness). No blocking issues — a clean, well-scoped implementation of #132's "freeze the debt, block new errors, ratchet later" plan, using standard tooling (`mypy-baseline`), mirroring the repo's existing CI-gate pattern, with **zero `src/` runtime changes** (verified — cannot affect the #156 goldens). Reviewers installed `mypy-baseline` and empirically confirmed the gate's key safety property: a mypy **crash does not silently pass** — the 329 baselined errors vanish from output, `filter` counts them as resolved, and CI fails (exit 100). Negative control works (a new untyped def → `new: 1` → CI fails); line-zeroing makes it robust to code movement; `tests/` is correctly excluded; ubuntu-only is right (no `sys.platform` branches in `src`). The items below are minor docs/hardening — none block merge.

## Important (worth addressing)

1. **Baseline is coupled to the locked toolchain, but the pin is loose and the docs overstate stability** (flagged independently by four of five reviewers). The 329-error baseline was generated against the locked **mypy 2.1.0** (and the inline types in the locked numpy/pandas), but `pyproject.toml` floors `mypy>=1.11.0` (a wide range), and `docs/CONTRIBUTING.md` claims "the baseline only ever shrinks" — which a `uv lock --upgrade` bumping mypy or a typed dep can falsify by *growing* it, red-walling an unrelated PR. Three cheap fixes:
   - add `--frozen`/`--locked` to the `type-check` job's `uv sync` so CI can't drift off the lock;
   - tighten the pin toward the locked major (`mypy>=2.1` or `mypy~=2.1`);
   - one sentence in CONTRIBUTING §5: the baseline is valid only against the locked toolchain, so bumping mypy/typed-deps requires `mypy-baseline sync` — and soften the "only ever shrinks" line.

2. **Two docs-accuracy overclaims** (empirically verified):
   - CONTRIBUTING/spec say new **public** functions must be typed, but `disallow_untyped_defs` applies to **all** defs incl. private — the gate is *stricter* than documented, so just drop "public" (or say "public or private").
   - The gate guarantees a signature *is annotated*, not that the types are *meaningful*: with `ignore_missing_imports=true`, a `def f(x: str) -> int:` whose body returns `pd.read_csv(...).shape[0]` (pandas → `Any`) passes. One sentence noting the launch gate checks annotation *presence*, not `Any`-freedom (a future `disallow_any_*` ratchet), keeps the guarantee from being oversold.

3. **Crash-safety becomes a hole at zero debt** (the sharpest finding). The crash→fail behavior relies entirely on the baseline being *non-empty*. Solid today (329 errors), but once the baseline is paid down to zero — the stated end goal — a mypy crash that emits nothing yields `new=0` → exit 0 → silent green. Worth a cheap explicit guard (capture mypy's own exit code; fail on `2`/fatal) as a tracked follow-up, not now.

## Suggestions

- The CI `paths:` trigger omits `.mypy-baseline.txt`, so a debt-paydown PR that only edits the baseline won't run the gate — add it to `paths:`.
- A fix-without-`sync` fails CI with a bare `fixed: N / new: 0`; add a one-line hint ("you resolved baselined errors → run `mypy-baseline sync`") to the job step or CONTRIBUTING so a confusing red becomes an obvious action.
- Windows-local repro of the documented command can hit `UnicodeEncodeError` on the colored summary glyph — note `NO_COLOR=1` / `PYTHONIOENCODING=utf-8` in CONTRIBUTING (CI is unaffected, ubuntu).
- `design.md`'s prose examples omit `--baseline-path` (they'd target the no-dot default file) — 2-line doc fix; all *operative* sites (CI + CONTRIBUTING commands) are correct.
- Nice tie-in: the baseline captures 22 real `files_generated … list[Path]` vs `list[str | Path]` (invariant) errors — the #157 cleanup would let those be `sync`'d out, so the two ratchets reinforce each other.

No blockers — approvable once you've eyeballed items 1–2 (both one-line fixes).

---

eberrigan (approved):

The single hardening commit (`1c500c38`) addresses every item from the review, and CI is fully green — including the `type-check` gate itself, so the tightened pin + `--frozen` still pass with the baseline unchanged.

- **Baseline/toolchain coupling** — `type-check` now installs with `uv sync --frozen --group dev`, and `mypy` is pinned `>=2.1,<3` (tracking the locked major) so a `uv lock --upgrade` can't silently drift the baseline. CONTRIBUTING now states the baseline is valid only against the locked toolchain and that bumping mypy/typed-deps can legitimately *grow* it → do the bump + `mypy-baseline sync` in one PR (the misleading "only ever shrinks" line is fixed).
- **Docs accuracy** — "public" dropped throughout (CONTRIBUTING + spec): `disallow_untyped_defs` applies to all defs, public or private. And the guarantee is now stated precisely — the gate checks a signature is *annotated*, not `Any`-free (a future `disallow_any_*` ratchet).
- **Suggestions** — `.mypy-baseline.txt` added to the CI `paths:` trigger (debt-paydown PRs run the gate); the `fixed: N / new: 0` → `mypy-baseline sync` hint and the Windows `NO_COLOR` note are in CONTRIBUTING; design.md prose uses `--baseline-path`; and the zero-debt crash-safety guard is recorded as a tracked follow-up.

Closes #132. Nice, thorough turnaround — ship it.

---

PR #159, opened 2026-06-17
refactor: centralize path serialization in pipeline (#157)

egao28:
Closes #157.

## Why

PR #156's cross-OS serialization round-trip gate caught a latent bug: pipeline steps hand-stringify paths with `str(path)` before they reach the central serializer, so on Windows a serialized path becomes `out\a.csv` instead of `out/a.csv`. #156 fixed only one site; the same anti-pattern was repeated across ~21 sites in the pipeline steps. Each pre-stringification **defeats** `convert_to_json_serializable`'s `Path → as_posix()` normalization and bakes in `\` on Windows, in the `files_generated` and `metadata` fields of `pipeline_summary.json` (the per-run provenance manifest) and in the standalone `*_manifest.json` files.

## What changed

- **Store bare `Path` at all producer sites** (13 step files) — let the serializers normalize once, centrally.
- **Tighten the type** `files_generated: List[str | Path]` → `List[Path]` in `core.py`, `summary.py`, `task.py` so the union can't re-invite divergence. (`List` is invariant, so this lands with the producer edits in one commit.)
- **Fixed a second serialization sink** the issue didn't call out: `BaseStep.save_json` used `json.dump(..., default=str)`, which would re-bake backslashes for a raw `Path` in the standalone manifests. The `default` hook now normalizes **only** `Path` via `as_posix()` and keeps `str()` for everything else — so the only behavior change is OS-independent paths (no numpy/datetime regression, which a full switch to `convert_to_json_serializable` would have risked against the golden-template tests).
- **Generalized** `convert_to_json_serializable`'s `isinstance(obj, Path)` → `PurePath` (a safe superset — every `Path` is a `PurePath`) so an OS-independent `PureWindowsPath` regression test is possible.

## Tests

- Fixed 4 existing assertions that did exact `str(x) in result.files_generated` membership (these break at runtime once elements are `Path`); they now compare on `Path`, which also enforces `Path` storage.
- Added an **OS-independent** cross-OS test using `PureWindowsPath` (the bug is invisible on Linux/macOS because `str(Path("out/a.csv"))` already yields `out/a.csv` there) plus a `None`-path assertion, to `tests/test_result_serialization.py`.
- Added `tests/test_no_path_prestringify.py` — an **AST source guard** (multi-line aware, unlike a grep) that fails if a `str(path)` is reintroduced into a step's `files_generated`/`metadata`.

## Docs

CHANGELOG `Fixed` entry, the producer-side "store `Path`, never `str(path)`" rule in `docs/reproducibility.md` (the gate's named source of truth), a pointer in `openspec/project.md`, and an updated `summary.py` comment reflecting the now-systemic contract.

## Verification

- `tests/test_result_serialization.py` (the cross-OS `serialization-gate`) + the new guard green.
- All directly-affected suites green: cross-platform steps, trait-redundancy, dashboards/interactive/static, heritability, merge, depth profiles, provenance, golden templates, reproduction fixtures.
- `ruff` + `black` clean; `mypy` shows **no new errors** from the type change (only pre-existing pandas-stub/`image_links` baseline entries).
- Confirmed no committed golden fixture asserts on or carries backslash serialized paths, so the #146 reproduction fixtures are undisturbed.

## ⚠️ Merge-order note

The mypy baseline gate is on the still-open #158, not `main`. **Land this before #158** (its baseline then freezes with the 19 `files_generated` invariance entries already resolved). If #158 merges first, resync `.mypy-baseline.txt` in this PR — CI runs `mypy-baseline filter` without `--allow-unsynced`, so a stale baseline fails the gate.

This PR follows an OpenSpec change (`openspec/changes/refactor-path-serialization/`) reviewed via the review-openspec subagent team.

Related: #156, #133.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -66,6 +66,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `diagnose_heritability_issues`, and `compare_trait_heritabilities` (Part of #116).
 
 ### Fixed
+- Pipeline provenance manifests now emit forward-slash (POSIX) paths on every OS.
+  Steps previously hand-stringified paths with `str(path)` before serialization,
+  which baked in backslashes on Windows (e.g. `out\a.csv` instead of `out/a.csv`)
+  in the `files_generated` and `metadata` fields of `pipeline_summary.json` and in
+  the standalone `*_manifest.json` files. Producers now store `Path` and let the
+  central serializers normalize once via `Path.as_posix()` (`convert_to_json_serializable`
+  and the `save_json` default hook); the `files_generated` field type was tightened
+  to `List[Path]` so the divergence can't recur. Completes the single-site fix from
+  #156. (#157)
 - Add the missing `from typing import Any` import in `statistics.py` so
   `typing.get_type_hints()` no longer raises `NameError` on the three functions that
   annotate `Dict[str, Any]`, unblocking downstream tool-schema generation.
--- a/docs/reproducibility.md
+++ b/docs/reproducibility.md
@@ -119,6 +119,17 @@ intentionally asymmetric: `ndarray`→list, unknown objects→`"<TypeName>"`). `
 survives an in-process round-trip and is compared NaN-aware; note it is non-standard
 JSON, so a stricter external consumer may reject it.
 
+**Path normalization is centralized — store `Path`, never `str(path)`.** Producers
+(pipeline steps, result objects) hand a `Path` object to the serializer and let it
+normalize to a POSIX string (`Path.as_posix()`) exactly once. Pre-stringifying with
+`str(path)` defeats this and bakes in OS-specific separators (backslashes on Windows),
+producing `out\a.csv` instead of `out/a.csv` in `pipeline_summary.json` and the
+standalone `*_manifest.json` files (#157). Both serialization sinks own this:
+`convert_to_json_serializable` (for the summary) and the `BaseStep.save_json` `default`
+hook (for standalone manifests). The `StepSummary.files_generated` field is typed
+`List[Path]` so the rule can't silently regress; [`tests/test_no_path_prestringify.py`](../tests/test_no_path_prestringify.py)
+is an AST guard that fails if a `str(path)` is reintroduced into a step.
+
 ## CI enforcement
 
 The gates run on every pull request and can be set as required status checks in
--- a/openspec/changes/refactor-path-serialization/proposal.md
+++ b/openspec/changes/refactor-path-serialization/proposal.md
@@ -1,0 +1,24 @@
+## Why
+
+PR #156's cross-OS serialization round-trip gate caught a latent bug: pipeline steps hand-stringify paths with `str(path)` before they reach the central serializer, so on Windows a serialized path becomes `out\a.csv` instead of `out/a.csv`. #156 fixed only one site (`PipelineSummary.to_json` no longer pre-`str()`s `files_generated`).
+
+The same `str(path)` anti-pattern is repeated across ~20 sites in the pipeline steps (issue #157). Each pre-stringification **defeats** `convert_to_json_serializable`'s `Path → as_posix()` normalization and bakes in `\` on Windows. The union type `files_generated: List[str | Path]` is what invites this divergence: it lets each producer choose to store a `str` or a `Path`. This is the per-run provenance manifest that rolls up into `pipeline_summary.json`, so the divergence is observable in shipped artifacts.
+
+## What Changes
+
+- Stop hand-stringifying paths at the producers across the pipeline steps. Store bare `Path` and let `convert_to_json_serializable` normalize once, centrally (no per-producer `.as_posix()` — that just relocates the choice). Affected sites:
+  - `files_generated=[str(...)]` / `.append(str(...))`: `calculate_cross_platform_correlations.py`, `calculate_trait_enrichment.py`, `reduce_trait_redundancy.py` (×3), `visualize_cross_platform.py` (×4), `load_cross_platform_data.py` (×3)
+  - `metadata[...] = str(path)` (and `[str(f.relative_to(run_dir)) …]`): `merge_all_traits.py`, `visualize_depth_profiles.py`, `filter_heritability.py`, `generate_dashboards.py`, `generate_summary_viz.py`, `load_above_ground.py`, `generate_interactive.py`, `generate_static_figures.py`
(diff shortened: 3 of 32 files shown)


eberrigan (changes requested):
## Review Summary

The refactor is directionally right and the core `pipeline_summary.json` path (`to_json` → `convert_to_json_serializable`, now `PurePath`-aware) is verified backslash-free on Windows. Consumers were traced — none do string ops on the now-`Path` elements, aggregation preserves type, `None`→`null`, and no numeric/golden result changes (the goldens don't assert on paths). But CI is red on a real bug, and — confirmed against committed golden manifests — the PR does **not** fully deliver its "normalize in exactly one place / cross-OS-stable manifest" goal: two manifest fields remain backslashed on Windows.

## Blocking Issues

**1. Windows CI fails — `tests/test_no_path_prestringify.py:64` reads source without an encoding.** `ast.parse(path.read_text(), ...)` defaults to cp1252 on Windows and raises `UnicodeDecodeError: 'charmap' codec can't decode byte 0x81` on `generate_dashboards.py` (emoji) and `visualize_cross_platform.py` (`ρ`). It's the only Windows failure (2/2). Fix: `path.read_text(encoding="utf-8")`. (Ironic for a cross-platform-hygiene PR — same family as the `as_posix()` rule: on Windows you must be explicit, here with `encoding="utf-8"`.)

**2. `output_directory` is still backslashed on Windows — `pipeline/pipelines/base_pipeline.py:98`.** `output_directory=str(self.run_dir)` pre-stringifies, and `PipelineSummary.output_directory` is typed `str` (`summary.py:68`), so `convert_to_json_serializable` can't re-normalize it. Confirmed in a committed golden: `tests/fixtures/real/wheat_edpie/expected/cross_platform/root_core_vs_cylinder/pipeline_summary.json:1352` has `"output_directory": "pipeline_runs\..."` while sibling `exp*_data_path` fields are forward-slash — i.e. the *same* manifest the PR is fixing stays internally inconsistent. The new AST guard only scans `pipeline/steps/`, so it misses this. Store the bare `Path` (and let the serializer normalize), and widen the guard or add a manifest-level "no backslash" assertion.

**3. A third, unfixed sink — the viz `summary.json` — `pipeline/steps/generate_summary_viz.py`.** The producer was correctly changed to store `run_directory` as a `Path` (line ~56), but this step writes its own `summary.json` through a local `_make_json_serializable` that still does `return str(obj)` for `Path` (line ~328) and `json.dump` with no `default=` hook — so on Windows `run_directory` stays backslashed. Confirmed in `tests/fixtures/real/wheat_edpie/expected/viz/cylinder/summary.json:4` (`run_directory` backslash, `data_file` forward-slash in the same file). The spec's ADDED requirement says normalization happens in "exactly one place" via `as_posix()` "on every OS" — this is a counterexample. Route it through `convert_to_json_serializable`, or change line ~328 to `obj.as_posix()` for `PurePath`, and regenerate that golden.

## Important Issues

- **Two serializers, divergent type predicates.** `convert_to_json_serializable` uses `isinstance(obj, PurePath)`; the new `save_json` hook (`core.py:131`) uses `isinstance(o, Path)`. A `PureWindowsPath` reaching `save_json` falls to `str()` → backslashes. Dormant today (producers create concrete `Path`), but it's a latent trap and untested — use `PurePath` in both (a shared `_path_to_posix` helper would fix it in one place).
- **The `save_json` second sink has no test.** The hook's `as_posix()` branch is never exercised — the existing `test_base_step_save_json` passes no `Path`, and the dashboard-manifest test only checks key presence, never reads the written JSON for forward slashes. Add a test that writes a manifest with a `Path` and asserts `/` (and no `\`).
- **`PipelineSummary.load()` doesn't rehydrate `List[Path]`** (`summary.py:115-121`) — `StepSummary(**step)` repopulates `files_generated` as `List[str]` despite the new annotation. Latent (no consumer calls `Path` methods on a loaded summary), worth a coercion or comment.
- **Missing-`encoding=` is a pre-existing pattern** at several JSON/markdown writers (`core.py:123` `open(... "w")` inside the very method this PR edits; `summary.py:104/117`; `data_utils.py:792`; `generate_summary_viz.py` writers). Not this PR's job to fix all, but fixing `core.py:123` here is reasonable; the rest → a tracked follow-up.

## Suggestions

- The AST guard is good and multi-line-aware, but foolable by `str` aliasing / f-strings / `os.fspath` / `"%s" %` (acceptable for catching *accidental* reintroduction — just soften the docstring claim), and slightly over-broad inside `files_generated` (flags any `str(...)`, e.g. `str(count)`).
- `STEPS_DIR.glob("*.py")` is non-recursive — a future `steps/<subdir>/*.py` is silently un-guarded; use `rglob` or assert no subdirs. Also guard against a vacuous pass: `assert len(_step_files()) >= 13`.
- The 4 changed assertions (`str(x) in files_generated` → `Path` comparison) are a genuine improvement — they now actually enforce `Path` storage.

The core change is sound; items 1–3 are what stand between this and an approve, and 2 & 3 are needed for it to truly close #157 (one normalization point) rather than fix the manifest's `files_generated`/`metadata` fields while leaving `output_directory` and the viz `summary.json` backslashed on Windows.

---

eberrigan (approved):

The single fix commit (`15a076f3`) resolves all three blocking items, and CI is now fully green — including `Tests (windows)` (21m), the job that was red. This now genuinely delivers #157's "normalize in exactly one place": every manifest path sink routes through `as_posix()`.

**Blockers — all fixed:**
- **Windows encoding crash** — `test_no_path_prestringify.py:64` now `path.read_text(encoding="utf-8")` (comment notes the non-ASCII emoji/`ρ`).
- **`output_directory`** — `base_pipeline.py` stores the bare `Path` and `PipelineSummary.output_directory` is retyped `str | Path`, so the serializer normalizes it (golden regenerated to forward-slash).
- **Third sink (viz `summary.json`)** — `generate_summary_viz.py` `_make_json_serializable` now normalizes `PurePath` via `as_posix()` (golden regenerated).

**Important items also addressed:**
- The two serializers are unified behind a single `path_to_posix` helper (`data_utils.py:56`), used as `save_json`'s `default=` hook — no more `Path` vs `PurePath` predicate split.
- `save_json` is now covered by `test_base_step_save_json_normalizes_paths_to_posix`, which round-trips a `PureWindowsPath` and asserts forward-slash output on any host.
- `PipelineSummary.load()` rehydrates serialized path strings back to `Path`.

**Golden regenerations are clean** — the only changes in the two committed manifests are the `output_directory` / `run_directory` backslash→forward-slash lines; no numeric or other field changed, so the #146 reproduction results are undisturbed.

Remaining items are all non-blocking suggestions (AST-guard foolability, non-recursive `glob`, empty-collection vacuous-pass guard) — optional hardening. Closes #157; nice, thorough turnaround.

---

PR #162, opened 2026-06-17
Add numerical-stability golden gate for pinned scientific deps

egao28:
## Why

Our deps are fully pinned via `uv.lock` against bleeding-edge scientific packages (numpy 2.x, pandas heading toward CoW-by-default, umap-learn riding the numba/BLAS JIT stack). The reproducibility gate we already ship — `tests/test_reproducibility.py` — is a **same-machine double-run**: it calls each stochastic function twice with `random_state=42` and asserts the two runs agree. That proves *determinism*, but it is structurally blind to *drift*: under a numba/numpy/pandas bump **both** runs move together and the gate still passes (`docs/reproducibility.md` says as much).

Nothing pinned our numerical output to a known-good **ground truth**. For UMAP the golden-fixture test asserts only embedding *shape and finiteness*, never coordinates; no golden cluster labels existed; pandas trait tables were checked only at the scalar-summary level. A dependency bump could silently rotate the UMAP manifold, reshuffle clusters, or flip a pandas groupby path with nobody noticing.

Mirrors upstream [Salk-Harnessing-Plants-Initiative/bloom#141](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/issues/141).

## What

A **golden-vs-committed numerical-stability drift gate** (`tests/test_numerical_stability.py`) that recomputes the UMAP, clustering, and pandas trait-aggregation paths on the `turface_19` reference slice and pins them to committed golden artifacts — complementing, not replacing, the determinism sweep. Tolerance-based, not bit-exact, with every threshold grounded in the dataset's **measured** same-stack spread:

- **UMAP embedding** — `scipy.spatial.procrustes` superimposition, then `np.allclose` on the **aligned coordinate matrices** (`atol=1e-6`). Aligned matrices, not the disparity scalar: measured same-stack spread ≈ `3e-17`, yet a real `1e-3` nudge yields ≈ `2.6e-5` — so `1e-6` is noise-immune yet sensitive.
- **Cluster labels** — `adjusted_rand_score > 0.95` (permutation-invariant) + the pinned `n_clusters=3` asserted (guards the clusterer's internal `len//10` clamp).
- **Trait summary** — `pd.testing.assert_frame_equal(rtol=1e-10)` on a per-genotype `groupby().agg(["mean","std"])` — the CoW-sensitive path.

**No new dependencies** — `scipy` and `scikit-learn` are already in `dependencies`.

### Single-OS by design
UMAP is not bit-reproducible across operating systems, and these tolerances are below the cross-OS float floor. The golden is generated on macOS and the test **self-skips elsewhere** (so the cross-platform `tests` matrix stays green), running in a dedicated single-OS `numerical-stability` CI job. `golden_provenance.json` records the OS / Python / dependency versions the golden was generated under, so staleness is a diff. Regenerate via `scripts/regenerate_numerical_stability_golden.py` on **major** numba/numpy/umap-learn/pandas bumps past tolerance only (reviewer-approved); never on patch bumps within tolerance.

## How it was built

Scaffolded as an OpenSpec change (`openspec/changes/add-numerical-stability-gate/`), critically reviewed with the 5-agent `review-openspec` skill, then revised before implementation. The review's blocking findings are all resolved in code:

- **Single-OS was aspirational → enforced** via module `skipif` + dedicated job (an unmarked test would otherwise have leaked onto the Windows/macOS matrix and failed at the tight tolerances).
- **Thresholds were copied from the issue → measured and derived**; the UMAP assertion compares aligned matrices because the disparity scalar is too insensitive (a 1e-3 nudge stays under `1e-6` on the scalar).
- **Failure modes specified**: missing input fails loudly with a regen pointer; new NaNs caught; negative controls prove the comparison bites *and* the recompute path is live (not golden-vs-golden).

## Test plan

Verified locally under py3.11 (matching CI):

- `tests/test_numerical_stability.py` — **11 passed** (4 golden assertions, 2 sensitivity, 3 negative-control/edge, 1 cluster-count, 1 provenance)
- `tests/test_reproducibility.py` — **33 passed** (no regression)
- `black --check`, `ruff check` — clean
- `openspec validate add-numerical-stability-gate --strict` — valid

**Note:** the golden was generated on local macOS/arm64; the `numerical-stability` CI job (macos-latest, arm64) is the first cross-machine confirmation. If it diverges past tolerance, the provenance record + one-command regen script make rebasing the golden trivial.

## Files

- `tests/test_numerical_stability.py` — the gate
- `tests/numerical_stability_recompute.py` — shared recompute logic (single source of truth)
- `scripts/regenerate_numerical_stability_golden.py` — regeneration + provenance
- `tests/fixtures/real/wheat_edpie/expected/numerical_stability/` — golden artifacts + provenance
- `tests/fixtures.py` — session-scoped golden loader
- `.github/workflows/ci.yml` — dedicated macOS gate job
- `docs/reproducibility.md`, `tests/fixtures/README.md`, `docs/CHANGELOG.md` — docs


eberrigan (commented):
## Review Summary

A genuinely strong, well-engineered drift gate that closes a real hole the same-machine determinism sweep can't (a dep bump moves both runs together and passes). The subagents verified the important things hold: **the gate actually executes in CI and bites** (`CANONICAL_PLATFORM="darwin"` matches the `numerical-stability` job's `runs-on: macos-latest`; it self-skips the cross-OS matrix correctly — not green-by-skip), negative controls genuinely fail (ARI≈0.16, embedding/trait drift), the recompute path is live (re-runs UMAP with `random_state=7` vs `42`, not golden-vs-golden), row-order alignment is correct (both sides sort by `ID_COL`), the recompute is seed-fixed so it isn't flaky, and the recompute logic is single-sourced. No new dependencies (lockfile untouched), the regen script is safe (manual-only, no subprocess/eval), goldens are plain data, and `golden_provenance.json` carries no secrets/paths.

**Verdict: no blocking issues; a handful of IMPORTANT robustness/scoping items worth addressing before merge.** Posting as a comment.

## Important

1. **The `n_clusters == 3` guard is inert for this dataset.** The clamp it claims to protect is `max(2, len(df)//10)` = 15 for the 153-row slice, and `N_CLUSTERS=3 < 15`, so the clamp can never fire here — `test_cluster_count_honored` asserts `3 == 3` where clamping is impossible. It only guards against a *shrunk* dataset, not a regression to the clamp formula (`len//50`, `len//100` would silently pass). Add a small-slice test (e.g. ~25 rows) asserting the clamp engages, or document that the guard's scope is "dataset didn't shrink."

2. **Procrustes masks similarity-transform drift, and there's no direct neighbour-structure check.** Confirmed empirically: a global uniform rescale (1.0001×) or rotation (1e-4 rad) is *fully* aligned away (≈4e-17), and `test_embedding_tolerates_rigid_transform` enshrines that. That's defensible (neighbour structure is what matters scientifically) — but the gate never *directly* verifies neighbour preservation, so a UMAP drift that is structurally a similarity transform reads as "no drift." Add a kNN-overlap / trustworthiness assertion, or explicitly document that similarity-transform-shaped drift is out of scope.

3. **`ARI > 0.95` is the issue's number, not a derived floor.** On 153 points, 0.95 tolerates ~5–8 reshuffled boundary points, so a real cluster-boundary drift could stay above it. Derive the floor from a boundary-perturbation sweep (as the embedding `atol` was derived), or document the measured headroom (same-stack ARI = 1.0).

4. **No automated staleness check on the golden vs the running stack.** `test_provenance_records_stack` only asserts the provenance fields are populated — nothing compares the recorded numpy/pandas/umap-learn/numba versions to the *running* ones. A silent dep bump within tolerance leaves the golden valid-by-construction and the provenance quietly aging; staleness is detectable only by manual `git diff`. Add a test that warns/fails when the running stack diverges from `golden_provenance.json`.

5. **The single-OS gap is real and under-surfaced for the deployment target.** The gate runs only on macOS/arm64 and self-skips on Linux/Windows — where this library actually runs (including bloom-mcp in Docker on Linux). A numerical regression that manifests only on Linux is **never caught**. The trade-off is reasonable (UMAP isn't cross-OS bit-reproducible, and most dep-bump drift is OS-independent), but `docs/reproducibility.md` frames it only as "the matrix stays green" — it should explicitly warn that deployment-OS-specific drift is out of this gate's scope.

6. **Shape-mismatch and new-NaN failures bypass the friendly "drifted + regenerate" message.** A dep bump that changes the embedding's row count surfaces as a raw `scipy` `ValueError: Input matrices must be of same shape` (and `adjusted_rand_score`'s "inconsistent numbers of samples"), not the regen pointer the PR promises for *all* failure modes; a new NaN in a future input → `recompute_labels` pairs `cluster_labels` (post-`dropna`) with the un-dropped `ID_COL` → opaque length-mismatch error. Add an explicit equal-length / equal-`ID_COL`-set precondition that raises with the regen pointer.

7. **Latent skew in `test_recompute_path_is_live_not_tautology`.** This anti-tautology guard re-implements `recompute_embedding`'s frame construction inline rather than calling the shared module — the one place the "single source of truth" leaks. If `recompute_embedding` changes (column order/dtype/id alignment), the golden and the real gate move together but this guard keeps the old shape, weakening the proof. Thread a `seed` parameter through the shared `recompute_embedding` and call it here.

## Suggestions

- **Missing *golden* artifact isn't tested** (only missing *input* is). A deleted golden raises a raw `FileNotFoundError` from `pd.read_csv`, not the regen-pointer message the spec promises — add the symmetric test.
- **Assert the provenance's recorded tolerances equal the live constants**, so editing `ATOL_PROCRUSTES`/`ARI_FLOOR`/`TRAIT_RTOL` without regenerating is flagged.
- **`align_labels` sorts each side independently** — assert the golden and recomputed `ID_COL` sets are identical before ARI (an equal-length-but-disjoint barcode set would mispair silently; not reachable today, but cheap to fence).
- **Sensitivity lower bound is only implicit:** there's a test that a 1e-3 nudge is *caught*, but none asserting same-stack noise *passes* below `atol` — the golden-vs-recompute pass covers it indirectly.
- **Supply-chain (repo-wide, pre-existing — not a regression):** GitHub Actions use floating tags (`@v4`, `@v6`) rather than commit SHAs, and there's no top-level `permissions:` block. The new job follows the existing convention, so it introduces no new exposure; consider SHA-pinning + `permissions: contents: read` repo-wide as separate hardening.

## Confirmed sound (credit where due)
Ground-truth-vs-characterization is correctly scoped and honestly claimed (a *drift detector*, not correctness validation — no overclaim); the `atol=1e-6` Procrustes derivation is sound and reproducible (a subagent reproduced the ~2.6e-5-on-1e-3-nudge vs ~3e-17 same-stack-spread numbers); `rtol=1e-10` on the trait `groupby` is justified (ddof=1 both sides, NaN positions checked, covered by a NaN test); CI is fully green including the new gate; TDD ordering in `tasks.md` is honest.

---
*Review by a 5-subagent team (Code Quality · Testing · Scientific Rigor & Reproducibility · Supply-Chain/CI Integrity · Behavioural Correctness), adapted from `/review-pr` for this scientific Python library and verified against the cloned source. No blocking issues found.*

eberrigan (approved):
## Review Summary — APPROVE ✅

Full subagent-team review (Code Quality · Testing · Statistical Rigor · Performance/Memory · Behavioural Correctness). **No blocking issues** — this is unusually rigorous, purely-additive test infrastructure (zero `src/` changes, verified). The team confirmed the two things that kill golden gates: it **recomputes live** (not golden-vs-golden — `test_recompute_path_is_live_not_tautology` proves it by re-running UMAP with seed 7 and asserting failure), and it **genuinely executes** (not skips) on the macOS job. Single source of truth shared between test and regen script; every tolerance a named constant with **measured in-code derivation**; Procrustes-align→`allclose(1e-6)` is statistically sound and the negative/sensitivity controls bite in both directions. Code quality 9/10. Approving; the items below are non-blocking hardening worth a follow-up pass before this becomes a required check.

## Important (non-blocking — recommend addressing soon)

1. **Pin the CI job to `macos-14`, not `macos-latest`.** The golden is arm64-pinned with tolerances explicitly below the cross-OS float floor. `macos-latest` == arm64 today (macos-15) so it matches, but it's a deliberately moving label (another image migration is slated for mid-2026), and the repo's own `tests` matrix already pins `macos-14`. A future re-point could red the gate for a non-drift reason. Highest-value change.
2. **Key on arch + assert provenance.** The gate skips on `sys.platform != "darwin"` (OS) but doesn't check `platform.machine()`, and `golden_provenance.json` is recorded but never asserted against the running env. Add a `platform.machine()` / `provenance["platform"]` check so provenance is load-bearing and a runner-image change self-explains instead of surfacing as a mystery tolerance failure.
3. **Add a no-skip tripwire.** A fully-skipped run is green and indistinguishable from a pass (`-rs` only reports skips). If `CANONICAL_PLATFORM` were mis-set or the platform string changed, all 11 tests skip → green no-op (the spec's "cannot pass vacuously" concern). Add a non-skipped sentinel asserting it ran on the canonical OS, or a CI step that fails if 0 tests collected.
4. **Missing/truncated *golden* file fails with a bare pandas/scipy error, not the spec-required regen pointer.** Only the missing-*input* path has the loud "run the regen script" message + a test; the golden loader in `tests/fixtures.py` does unguarded `read_csv`/`read_text`. Failure direction is safe (fails, doesn't pass), but add a guarded loader with the regen pointer + a `test_missing_golden_fails_loudly`.
5. **Doc/spec accuracy:** (a) the spec & `docs/reproducibility.md` say the gate "skips on macOS / matrix stays green," but it *also* runs in the `tests` matrix macOS (macos-14) leg, so a drift reddens two jobs and the "matrix green" guarantee only holds for Ubuntu/Windows; (b) document that UMAP `atol=1e-6` is **post-Procrustes-normalization** (unit-norm), not raw coords — the math is correctly calibrated (catches a ~1e-4–1e-5 raw nudge), just label it; (c) ARI `> 0.95` "leaves headroom" understates it — same-stack is exactly 1.0 and the floor only tolerates ≤2 of 153 relabeled (3 flips → fail); state the "≤2 samples" meaning or tighten toward `== 1.0`. `design.md` also still says `linux`/`ubuntu` (stale vs shipped `darwin`).

## Suggestions

- `test_cluster_count_honored` is tautological at n=153 (clamp=15, k=3 can't fire) — it's a tripwire for a future slice/clamp change, not an active assertion; note it.
- An embedding NaN fails via an *uncaught* `procrustes ValueError` (not the named "embedding drifted" `AssertionError`) and isn't tested — add a case / wrap for a clean message.
- The lossless CSV round-trip holds *only* because `recompute_embedding` widens UMAP's float32→float64 before `to_csv` — add a comment protecting that invariant (a float32 "optimization" would truncate to ~8 digits and erode the headroom).
- Pin & stamp the UMAP `n_neighbors`/`min_dist` (currently library defaults, not in provenance) so a future default change shows as a provenance diff.
- `encoding="utf-8"` on the two provenance `read_text`/`write_text` calls (ASCII today, cheap, matches the lab convention from #159).

Excellent, well-controlled gate — approving on the strength of the live-recompute proof and the measured-tolerance discipline. The runner-pin (item 1) and the no-skip tripwire (item 3) are the two I'd most encourage as quick follow-ups since the PR notes this CI run is the first cross-machine confirmation.

---

PR #166, opened 2026-06-23
feat: add public clean_traits_for_analysis minimal-QC entry point (#164)

egao28:
## Summary

Closes part of #164. Adds a single public, tested entry point —
`clean_traits_for_analysis` — that turns a raw wide trait table into a **clean,
analysis-ready** frame so downstream consumers (bloom-mcp Tier 3 `qc_clean` →
`pca_analysis`, notebooks) can run PCA/UMAP/clustering **without
`perform_pca_analysis` silently `dropna()`-ing every NaN row**
([`pca.py:775`](../blob/main/src/sleap_roots_analyze/pca.py#L775)).

It composes existing QC logic — no new cleanup algorithm:
1. resolve trait columns via `get_trait_columns` (if not passed),
2. run `apply_data_cleanup_filters` (the step-02 smart cleanup that drops bad
   **traits** first, then NaN rows, to minimize sample loss),
3. validate the result is analysis-ready,

and returns `(clean_df, trait_cols, cleanup_log)`.

## Single source of truth (the #116 pattern)

Rather than re-implement step 02/03 logic, this **exposes** the functions the QC
pipeline already uses and has the entry point **import** them:

- **Step 02:** `apply_data_cleanup_filters` is now exported in `__all__`.
- **Step 03:** its previously **inline** NaN check is extracted into importable
  `build_clean_validation_report` + `validate_clean_traits` (sharing one error
  formatter). `ValidateCleanStep` now calls these — a **transparent refactor**:
  same report, same saved `03_validation_report.json`, same `StepResult.metadata`,
  byte-for-byte same error message.

So the pipeline and the entry point share the same code and **cannot drift** at
the algorithm/validation level.

## Analysis-readiness validation

Runs in a **fixed order**, each raising a distinct, actionable `ValueError`:

1. **empty input** — raises its own message *before* delegating;
2. **no NaN** in surviving traits (via the shared `validate_clean_traits`);
3. **≥2 surviving samples** (message names the count);
4. **≥1 non-constant numeric trait** — `var(ddof=0) > 0`, the **same** basis
   `standardize_data` uses ([`pca.py:643-645`](../blob/main/src/sleap_roots_analyze/pca.py#L643)),
   so the gate and downstream agree.

`cleanup_log` is enriched with `effective_thresholds` and `validation_summary`
for auditability.

## Important behavioral note (defaults)

The entry point uses `apply_data_cleanup_filters`' **own** defaults
(`max_zeros_per_trait=0.5`, `max_nans_per_trait=0.3`, `max_nans_per_sample=0.2`,
`min_samples_per_trait=10`), which **differ from the QC pipeline's config
defaults**. Producing pipeline-identical output requires passing matched
thresholds. This is documented in the docstring, the spec, and recorded in
`cleanup_log["effective_thresholds"]`. "≥2 samples" is a *runnability* floor, not
a guarantee of scientific adequacy.

## Out of scope (per #164)

- The full `QCPipeline`.
- **Outlier detection/removal** (steps 05–07) → follow-up #165 (PCA runs fine with
  outliers; they are not NaNs).
- Stats / heritability / summaries; root-core / depth.
- Column-level zero-variance loss inside PCA (a quality concern; the ≥1-non-constant
  gate guarantees runnability, not per-trait standardization safety).
- `sanitize_trait_names` (pipeline-only preprocessing) — this is *why* entry-point
  and pipeline outputs can differ on raw input.

## Changes

| Area | Change |
|---|---|
| `src/.../data_cleanup.py` | `clean_traits_for_analysis`, `validate_clean_traits`, `build_clean_validation_report`, shared error formatter; deterministic-ordering comment |
| `src/.../pipeline/steps/validate_clean.py` | `ValidateCleanStep` now calls the extracted functions (transparent) |
| `src/.../__init__.py` | all four functions imported + in `__all__` |
| `tests/test_clean_traits_entry_point.py` | 24 new tests |
| `docs/API.md`, `docs/CHANGELOG.md` | documented |
| `openspec/changes/add-clean-traits-entry-point/` | proposal / design / spec / tasks |

## Testing

- **Full suite: 2333 passed, 19 skipped** (`uv run pytest`).
- New tests cover: no-NaN output + PCA row-`dropna` no-op; **sample loss minimized
  vs naive `dropna()`**; the four ordered validation errors + a passing
  mixed-constant case; kwargs pass-through; default column names + optional
  replicate; public-API surface (`__all__`, importability, identity, resolvable
  type hints); and the **byte-exact** shared validation message.
- **No behavior change to QC steps 01–03** verified by the unchanged
  `test_step_validate_clean` / `test_qc_pipeline` regression suites.
- `black --check`, `ruff check`, and `openspec validate --strict` all clean.

## Acceptance (#164)

- [x] `clean_traits_for_analysis` **and** `apply_data_cleanup_filters` in public `__all__`
- [x] No NaNs in output; sample loss minimized vs `dropna()`; validate raises on
      empty / all-NaN / single-constant; `perform_pca_analysis` dropna is a no-op
- [x] No behavior change to `QCPipeline` steps 01–03; deterministic; suite green

## Release coupling

Per #163, the `0.1.0a3` cut **does not currently include #164** — this must merge
before that release runs, or it slips to `0.1.0a4`. Needed to unblock bloom-mcp
Tier 3.

### DIFF
--- a/.mypy-baseline.txt
+++ b/.mypy-baseline.txt
@@ -1,83 +1,4 @@
-src/sleap_roots_analyze/pipeline/steps/validate_clean.py:0: error: Function is missing a return type annotation  [no-untyped-def]
-src/sleap_roots_analyze/pipeline/steps/validate_clean.py:0: note: Use "-> None" if function does not return a value
-src/sleap_roots_analyze/pipeline/steps/validate_clean.py:0: error: Item "None" of "StepResult | None" has no attribute "metadata"  [union-attr]
-src/sleap_roots_analyze/pipeline/steps/validate_clean.py:0: error: Argument "files_generated" to "StepResult" has incompatible type "list[Path]"; expected "list[str | Path]"  [arg-type]
-src/sleap_roots_analyze/pipeline/steps/validate_clean.py:0: note: "list" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
-src/sleap_roots_analyze/pipeline/steps/validate_clean.py:0: note: Consider using "Sequence" instead, which is covariant
-src/sleap_roots_analyze/pipeline/steps/transform_depth_data.py:0: error: Function is missing a return type annotation  [no-untyped-def]
-src/sleap_roots_analyze/pipeline/steps/transform_depth_data.py:0: note: Use "-> None" if function does not return a value
-src/sleap_roots_analyze/pipeline/steps/transform_depth_data.py:0: error: Argument "files_generated" to "StepResult" has incompatible type "list[Path]"; expected "list[str | Path]"  [arg-type]
-src/sleap_roots_analyze/pipeline/steps/transform_depth_data.py:0: note: "list" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
-src/sleap_roots_analyze/pipeline/steps/transform_depth_data.py:0: note: Consider using "Sequence" instead, which is covariant
-src/sleap_roots_analyze/pipeline/steps/transform_depth_data.py:0: error: Incompatible return value type (got "None", expected "float")  [return-value]
-src/sleap_roots_analyze/pipeline/steps/reshape_for_trait_qc.py:0: error: Function is missing a return type annotation  [no-untyped-def]
-src/sleap_roots_analyze/pipeline/steps/reshape_for_trait_qc.py:0: note: Use "-> None" if function does not return a value
-src/sleap_roots_analyze/pipeline/steps/reshape_for_trait_qc.py:0: error: Need type annotation for "metadata"  [var-annotated]
-src/sleap_roots_analyze/pipeline/steps/reshape_for_trait_qc.py:0: error: Incompatible types in assignment (expression has type "dict[Any, Any]", target has type "list[Any]")  [assignment]
-src/sleap_roots_analyze/pipeline/steps/reshape_for_trait_qc.py:0: error: Argument "files_generated" to "StepResult" has incompatible type "list[Path]"; expected "list[str | Path]"  [arg-type]
-src/sleap_roots_analyze/pipeline/steps/reshape_for_trait_qc.py:0: note: "list" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
-src/sleap_roots_analyze/pipeline/steps/reshape_for_trait_qc.py:0: note: Consider using "Sequence" instead, which is covariant
-src/sleap_roots_analyze/pipeline/steps/remove_outliers.py:0: error: Function is missing a return type annotation  [no-untyped-def]
-src/sleap_roots_analyze/pipeline/steps/remove_outliers.py:0: note: Use "-> None" if function does not return a value
-src/sleap_roots_analyze/pipeline/steps/remove_outliers.py:0: error: Item "None" of "StepResult | None" has no attribute "metadata"  [union-attr]
-src/sleap_roots_analyze/pipeline/steps/remove_outliers.py:0: error: Item "None" of "StepResult | None" has no attribute "metadata"  [union-attr]
-src/sleap_roots_analyze/pipeline/steps/remove_outliers.py:0: error: Item "None" of "StepResult | None" has no attribute "metadata"  [union-attr]
-src/sleap_roots_analyze/pipeline/steps/remove_outliers.py:0: error: Item "None" of "StepResult | None" has no attribute "metadata"  [union-attr]
-src/sleap_roots_analyze/pipeline/steps/remove_outliers.py:0: error: Item "None" of "StepResult | None" has no attribute "metadata"  [union-attr]
-src/sleap_roots_analyze/pipeline/steps/remove_outliers.py:0: error: Item "None" of "StepResult | None" has no attribute "metadata"  [union-attr]
-src/sleap_roots_analyze/pipeline/steps/remove_outliers.py:0: error: Item "None" of "StepResult | None" has no attribute "metadata"  [union-attr]
-src/sleap_roots_analyze/pipeline/steps/remove_outliers.py:0: error: Argument "files_generated" to "StepResult" has incompatible type "list[Path]"; expected "list[str | Path]"  [arg-type]
-src/sleap_roots_analyze/pipeline/steps/remove_outliers.py:0: note: "list" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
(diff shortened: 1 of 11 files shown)


eberrigan (changes requested):
## Review Summary

Full subagent-team review (Code Quality · Testing · Statistical Rigor · Performance/Memory · Behavioural Correctness). The design is clean, the single-source reuse is real, and the `ValidateCleanStep` refactor is genuinely byte-for-byte transparent (two reviewers confirmed identical report dict/keys/order, error string, metadata, and save-before-raise). But there are two blocking items — a red CI gate and a real default-behavior defect — plus several important issues.

## Blocking Issues

**1. The mypy baseline gate (CI) is RED.** `mypy-baseline filter` reports **4 new errors + 63 stale baseline entries**, and CI runs `filter` without `--allow-unsynced`, so both fail it:
- `data_cleanup.py:887` — **this PR's own** `clean_traits_for_analysis` has an unannotated `**cleanup_kwargs` (and `Any` is not imported). The gate correctly catching #166's code. Fix: `**cleanup_kwargs: Any` + add `Any` to the typing import.
- `data_utils.py:74` (`path_to_posix` untyped) and `reduce_trait_redundancy.py:180/284` (`list[str]` where `files_generated` is now `List[Path]` — a real residual type-lie from #159's sweep).
- `fixed: 63` means the committed `.mypy-baseline.txt` is stale (fallout from #159/#162 merging without `mypy-baseline sync`). To green the gate: annotate the new function, fix/sync the inherited errors, and `mypy-baseline sync`. Worth confirming whether `main` itself is currently mypy-red — if so this is partly a repo-health item #166 inherits, not solely its fault.

**2. With its own default thresholds, the entry point RAISES instead of returning an analysis-ready frame** (empirically reproduced). Defaults `max_nans_per_sample=0.2` / `max_nans_per_trait=0.3` *retain* samples/traits that still contain NaN, but gate-2 forbids any residual NaN → `ValueError`. A benign 12-sample × 5-trait frame with a single NaN (per-sample NaN fraction = 0.2, exactly the retain boundary) raises. So the public entry point, called with defaults on ordinary sparse-missing wide data, throws rather than delivering its "clean frame so PCA won't silently dropna" guarantee — and the docstring's "drops bad traits, *then NaN rows*" is misleading (it drops NaN-*heavy* samples, not the residual-NaN rows). It fails safe (raises, never returns dirty data), so it's not silent corruption — but it's a genuine default-behavior defect. Fix (pick one, documented): drop remaining NaN rows in surviving traits before gate-2 (matches the documented intent), default `max_nans_per_sample=0.0`, or make gate-2 a final cleanup pass.

## Important Issues

- **"Pipeline-equivalent" is unachievable even with matched thresholds.** `CleanupTraitsStep` also runs `sanitize_trait_names` (abbreviate / custom replacements / depth mapping) + column reorder before cleanup; the entry point does none of it, so matched thresholds still yield different trait names/identities/order than `02_data_samples_cleaned.csv`. The docstring should state name-sanitization is not applied and byte-equivalence with the pipeline is not a goal (publishable-results hazard).
- **The divergence is silent at runtime** — no warning/log; programmatic consumers (bloom-mcp `qc_clean`) never see the differing-defaults note that lives only in docstring prose. Surface the effective thresholds via `logger`/`warnings`.
- **The `≥2 samples` floor returns a statistically meaningless frame with no warning.** Consider warning in the p≫n regime (`n_samples < n_surviving_traits`).
- **Misuse → opaque errors:** an explicit `trait_cols` name not in `df` raises a bare pandas `KeyError`; duplicate column names degrade to the misleading "no trait columns found" message. Validate `set(trait_cols) <= set(df.columns)` (and detect duplicates) up front with an actionable `ValueError`.
- **Testing gaps:** tasks.md §2.12 is checked but the promised pipeline integration regression test does not exist (`tests/test_qc_pipeline.py` is untouched) — the transparent-refactor net is unit-level only; and the "cleanup reduces survivors to <2 samples" path is untested (the existing test pre-shrinks the input rather than exercising the cleanup path). The task list over-claims delivered coverage.

## Suggestions

- Threshold defaults (`0.5/0.3/0.2/10`) are hardcoded copies of `apply_data_cleanup_filters`' signature defaults — a drift seam in a PR whose purpose is preventing drift; derive from `inspect.signature(...)` or have the helper record its own effective thresholds.
- A non-numeric column passed explicitly in `trait_cols` survives validation (gate-4 excludes it via `select_dtypes`) but will break PCA later — consider a numeric-dtype check on explicit `trait_cols`.
- Bare `Dict` return type → `Dict[str, Any]`; consider a named `MIN_SAMPLES_FOR_ANALYSIS = 2`.

## Verified good (credit)
- The `ValidateCleanStep` refactor is byte-for-byte transparent; single-source reuse is real (no copy-pasted cleanup/validation logic).
- `var(ddof=0) > 0` is the correct constant-trait gate — it matches `standardize_data` exactly and is divide-by-zero-safe via sklearn's `_handle_zeros_in_scale` (a near-constant `var≈3e-31` trait passes and standardizes without inf/NaN). A tolerance would be *wrong* here.
- Stateless, vectorized, no double-copy introduced, deterministic; gate-2 genuinely guarantees no NaN silently reaches the returned frame.

---

eberrigan (approved):

The scoped defaults change (commit `2152d038`) is exactly what was requested, and CI is now fully green across all jobs (Lint, mac/ubuntu/windows Tests, all gates, mypy).

- **Entry point now defaults to the QC-canonical thresholds** — `_QC_DEFAULTS = {"max_nans_per_trait": 0.2, "max_nans_per_sample": 0.0}`, consulted before the signature default; the two that already match (`max_zeros_per_trait=0.5`, `min_samples_per_trait=10`) keep inheriting. So bloom-mcp/notebooks get QC-pipeline-aligned cleaning by default.
- **Scoped correctly** — the shared `apply_data_cleanup_filters` defaults are untouched (still `0.3`/`0.2`), so `visualization.py` and the `test_data_cleanup.py` suite are unaffected; the broader shared-default alignment stays with #167.
- **Docstring + tests + OpenSpec docs all updated** — the `effective_thresholds` assertion now expects `0.2`/`0.0`, the `quarter_nan_data` test verifies the 25%-NaN trait is dropped at the `0.2` default and kept only when a caller loosens it (proving caller overrides still work), and stale `0.3` comments are corrected.

Combined with the prior commit (both original blockers fixed — default-behavior defect + mypy gate — and every important/suggestion item addressed), this is complete. Closes part of #164. The shared-default change (#167) and the inherited mypy repo-health items remain tracked follow-ups, neither blocking here. Nice work.

---

PR #171, opened 2026-06-29
fix(qc): align cleanup defaults with canonical config

egao28:
Aligns the shared data-cleanup thresholds with the canonical QC config so every default-using caller cleans the same way the pipeline does. Closes #167.

## Changes
- **`apply_data_cleanup_filters` now uses the `CleanupConfig` defaults.** Signature defaults changed `max_nans_per_trait` 0.3 → 0.2 and `max_nans_per_sample` 0.2 → 0.0 (the other two already matched), so they equal `CleanupConfig()` with the `max_nan_fraction` ↔ `max_nans_per_sample` name mapping.
- **`clean_traits_for_analysis` inherits from the helper signature instead of carrying `_QC_DEFAULTS`.** Dropped the hardcoded `_QC_DEFAULTS = {0.2, 0.0}` copy; it now resolves all four thresholds from `inspect.signature(apply_data_cleanup_filters)`, giving a single source of truth. Behavior is unchanged (still 0.2/0.0).
- **`visualization.py` stale `0.3` fallbacks aligned to `0.2`.** In `create_trait_eda_plots`, the NaN threshold reference line and the simulate-branch `apply_data_cleanup_filters` call now fall back to the canonical 0.2, and `max_nans_per_sample` is passed explicitly (canonical 0.0) instead of relying on the function default.
- **Drift-guard tests prevent future mismatch.** New `TestCanonicalDefaultDriftGuard` asserts both `apply_data_cleanup_filters`'s signature defaults and `clean_traits_for_analysis`'s recorded effective thresholds equal `CleanupConfig()`. One existing test that implicitly relied on the old 0.3 default was made explicit.

## Caller audit (all 3 callers of `apply_data_cleanup_filters`)
- `CleanupTraitsStep` (pipeline) — passes explicit `config.cleanup.*`; unaffected.
- `clean_traits_for_analysis` — now inherits the canonical defaults; behavior unchanged.
- `create_trait_eda_plots` (visualization) — the only caller hard-coding the old 0.3 / implicitly relying on the old per-sample default; now aligned.

## Validation
- Affected suites pass: `test_data_cleanup`, `test_clean_traits_entry_point`, `test_step_cleanup`, `test_step_exploratory_analysis`, `test_visualization`, `test_qc_pipeline`, `test_pipeline_runner_summary`, `test_pipeline_summary_completeness`.
- `ruff check` passes on all changed files.

## Scope
No bloom-mcp changes, no release/version-pin changes, no unrelated reformatting.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,6 +7,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- `apply_data_cleanup_filters` bare-default thresholds tightened to the canonical
+  QC values — `max_nans_per_trait` `0.3 → 0.2` and `max_nans_per_sample`
+  `0.2 → 0.0` — so they equal `CleanupConfig()`'s defaults (with `max_nan_fraction`
+  ↔ `max_nans_per_sample` name mapping). The function signature is now the single
+  source of truth for canonical cleanup, and `clean_traits_for_analysis` inherits
+  it instead of carrying a hardcoded copy (#167). **Behavior note:** code calling
+  `apply_data_cleanup_filters` with the *bare* defaults now cleans more strictly —
+  `max_nans_per_sample=0.0` drops every sample that still has any NaN in a surviving
+  trait, which can remove more plants than before. The QC pipeline
+  (`CleanupTraitsStep` passes explicit `config.cleanup.*`) and the
+  `clean_traits_for_analysis` entry point (already pinned to `0.2`/`0.0` since #164)
+  are reproducibility-neutral — their effective thresholds are unchanged.
+
 ## [0.1.0a3] - 2026-06-24 (Pre-release)
 
 ### Added
--- a/src/sleap_roots_analyze/data_cleanup.py
+++ b/src/sleap_roots_analyze/data_cleanup.py
@@ -650,8 +650,8 @@ def apply_data_cleanup_filters(
     df: pd.DataFrame,
     trait_cols: List[str],
     max_zeros_per_trait: float = 0.5,
-    max_nans_per_trait: float = 0.3,
-    max_nans_per_sample: float = 0.2,
+    max_nans_per_trait: float = 0.2,
+    max_nans_per_sample: float = 0.0,
     min_samples_per_trait: int = 10,
     barcode_col: str = "Barcode",
     genotype_col: str = "geno",
@@ -665,13 +665,25 @@ def apply_data_cleanup_filters(
     3. Remove samples with many NaNs
     4. Remove traits with insufficient samples
 
+    The signature defaults are the **QC pipeline's canonical** cleanup thresholds:
+    they equal ``CleanupConfig()``'s defaults (with ``max_nans_per_sample`` mapping
+    to ``CleanupConfig.max_nan_fraction``), so every default-using caller cleans the
+    same way the pipeline does. A drift guard
+    (``tests/test_data_cleanup.py::TestCanonicalDefaultDriftGuard``) asserts this
+    equality so the two cannot silently diverge (#167).
+
     Args:
         df: Original dataframe
         trait_cols: List of trait column names
-        max_zeros_per_trait: Maximum fraction of zeros allowed per trait (0-1)
-        max_nans_per_trait: Maximum fraction of NaNs allowed per trait (0-1)
-        max_nans_per_sample: Maximum fraction of NaNs allowed per sample (0-1)
-        min_samples_per_trait: Minimum number of valid samples required per trait
+        max_zeros_per_trait: Maximum fraction of zeros allowed per trait (0-1).
+            Default ``0.5`` (canonical QC).
+        max_nans_per_trait: Maximum fraction of NaNs allowed per trait (0-1).
+            Default ``0.2`` (canonical QC; drops NaN-heavier traits sooner).
+        max_nans_per_sample: Maximum fraction of NaNs allowed per sample (0-1).
+            Default ``0.0`` (canonical QC; drops any sample that still has *any*
+            NaN in a surviving trait).
+        min_samples_per_trait: Minimum number of valid samples required per trait.
+            Default ``10`` (canonical QC).
         barcode_col: Name of the barcode/plant ID column (default: "Barcode")
         genotype_col: Name of the genotype column (default: "geno")
         replicate_col: Name of the replicate column if present (default: "rep")
@@ -925,10 +937,10 @@ def clean_traits_for_analysis(
           (``max_zeros_per_trait=0.5``, ``max_nans_per_trait=0.2``,
           ``max_nans_per_sample=0.0``, ``min_samples_per_trait=10``), so the
           analysis-ready frame matches what the QC pipeline produces rather than a
-          looser clean. (``apply_data_cleanup_filters``'s own signature defaults are
-          looser for two of these — ``max_nans_per_trait=0.3``,
-          ``max_nans_per_sample=0.2``; aligning them is tracked in #167.) Caller
-          kwargs override; the effective thresholds are recorded in
+          looser clean. These are inherited directly from
+          ``apply_data_cleanup_filters``'s signature defaults, which were aligned to
+          the canonical QC values in #167 (so there is no separate copy to drift).
+          Caller kwargs override; the effective thresholds are recorded in
           ``cleanup_log["effective_thresholds"]`` and logged at INFO.
         - This entry point does **NOT** apply trait-name sanitization
           (``sanitize_trait_names``) or column reordering that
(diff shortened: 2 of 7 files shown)


eberrigan (commented):
> **Verdict: COMMENT**

## Review Summary

Tight, well-motivated, single-purpose change. It removes a real duplication (the hardcoded `_QC_DEFAULTS` copy in `clean_traits_for_analysis`), makes `apply_data_cleanup_filters`'s signature the single source of truth for the canonical QC thresholds, fixes stale `0.3` fallbacks in `create_trait_eda_plots`, and adds a genuinely strong `TestCanonicalDefaultDriftGuard`. CI is green on ubuntu/windows/mac (3.11). **No blocking issues.** I verified the central reproducibility claim empirically and it holds: the v0.1.0a3 entry point already shipped `_QC_DEFAULTS = {0.2, 0.0}`, so `clean_traits_for_analysis`'s effective thresholds are byte-identical before/after (`0.5/0.2/0.0/10`) — the refactor is provably behavior-preserving, and the pipeline (`CleanupTraitsStep` passes explicit `config.cleanup.*`) is unaffected. The only true new behavior is the documented tightening of `apply_data_cleanup_filters`'s **bare** defaults (intended — the whole point of #167).

Credit where due: the drift guard's design (config-equality **plus** a literal-value pin) correctly defeats an in-tandem edit of both layers; and the regression-test audit was right — the one existing test whose behavior genuinely flips at 0.2 (`test_remove_nan_samples_max_nan_fraction_zero`, 0.25-NaN fixture) was surgically fixed with an explanatory comment.

## Important (non-blocking)

1. **No CHANGELOG entry for a public-API behavior change.** `apply_data_cleanup_filters` is exported in `__all__`, and its bare defaults tightened (`max_nans_per_trait` 0.3→0.2, `max_nans_per_sample` 0.2→0.0). Any external caller using bare defaults now cleans more strictly — `max_nans_per_sample=0.0` drops every sample with *any* residual NaN, which can drop far more plants than before. The entry point is reproducibility-neutral, but the low-level function isn't. Please add a `Changed` line under `## [Unreleased]` in `docs/CHANGELOG.md` as the reproducibility breadcrumb for downstream consumers (e.g. bloom-mcp).

2. **Stale `CleanupConfig` docstring contradicts the value this PR canonicalizes** — `pipeline/config/components.py:60`. The attribute is `max_nan_fraction: float = 0.0`, but its docstring still says `Recommended: 0.25`. This PR's premise (and `TestCanonicalDefaultDriftGuard`) elevates `CleanupConfig()` to "the single source of truth," yet a reader copying the recommended `0.25` gets the *opposite* of canonical (looser). The guard pins the literal value but can't see docstring prose. (Pre-existing, but the PR makes it glaring; the `max_nans_per_trait` "Recommended: 0.2-0.3" line is worth tightening to `0.2` for consistency too.)

## Suggestions

3. **The viz change is effectively untested.** The `0.3→0.2` axhline fallbacks and the new `max_nans_per_sample=0.0` simulate-branch pass-through in `create_trait_eda_plots` only execute when the `thresholds` dict omits those keys — but `viz_eda_thresholds` (`tests/fixtures.py:1722`) still passes `nan: 0.3` explicitly and no test exercises a key-less dict with `cleanup_log=None`. Update the fixture to `0.2` and/or add one test hitting the fallback path. Note also a silent diagnostic shift: a standalone caller passing an old 3-key dict (`nan`/`zero`/`outlier`) with `cleanup_log=None` now previews at `sample=0.0` vs the old implicit `0.2` (no pipeline impact — the pipeline always passes `cleanup_log`).

4. **Public-API default inconsistency.** `remove_nan_samples` (also in `__all__`) still defaults `max_nan_fraction=0.2` while the orchestrator/entry-point per-sample default is now `0.0`. Harmless in practice (`apply_data_cleanup_filters` always passes explicit values down), but two public "max NaN per sample" defaults now disagree. Worth a one-line note on the helper, or aligning it to `0.0`.

5. **`inspect.signature` resolution is name-fragile.** `threshold_names` is a hardcoded tuple of parameter-name strings; renaming a signature parameter surfaces as a `KeyError` only at call time. The drift guard covers *value* drift, not *name* drift. Cheap hardening: `assert set(threshold_names) <= set(sig.parameters)` once, or just note it. (The reflection approach itself is the right call here — it's what kills the `_QC_DEFAULTS` copy.)

6. **Entry-point test is a recorded-thresholds proxy, not a frame-parity test** — `test_clean_traits_entry_point_defaults_match_cleanup_config`. It asserts only `cleanup_log["effective_thresholds"]` on a no-filtering fixture, so it proves the *recorded* thresholds inherit, not that the cleaned *frame* matches the pipeline. That's acceptable — the threshold-equality drift guard is the architecturally correct parity proof, since the issue's literal "same cleaned table" test is actually unsatisfiable (`CleanupTraitsStep` runs `sanitize_trait_names` and `clean_traits_for_analysis` adds an extra `dropna`, so the frames are deliberately not equivalent). Just consider softening the test's docstring claim ("Proves the public entry point inherits the corrected defaults") to match what it checks.

## Note on a finding I ran down and dismissed

An automated pass flagged two sibling regression tests (`..._removed_samples_is_independent_copy`, `..._max_nan_fraction_one_keeps_partial_nan`) as "hollowed out" by the stricter 0.2 default. I checked both against the base branch: their fixtures put traits at 50% NaN (exceeds *both* old 0.3 and new 0.2) and/or rely on `min_samples_per_trait=10` with 2 samples, so they were already vacuous independent of this change. Not introduced by this PR — no action needed here (though they'd be worth strengthening someday, separately).

---

eberrigan (changes requested):
## Re-review of `0a7e9781`

Thanks for the thorough turnaround, @egao28 — every point from the last round is addressed well: the CHANGELOG `Changed` entry (with the reproducibility-neutral note), the `remove_nan_samples` docstring note, the `inspect.signature` name-drift guard, the `CleanupConfig` docstring fix (canonical-default phrasing replacing the misleading "Recommended: 0.25"), the `viz_eda_thresholds` fixture `0.3→0.2` **plus** a genuine new fallback test, and the softened entry-point test docstring. The substance is all there.

**Lifting requires only fixing CI** — the response commit turned it red with two mechanical issues. Both are one-liners; no design concerns.

## Blocking (CI red)

1. **mypy fails** — `src/sleap_roots_analyze/data_cleanup.py:1042`. The new name-drift guard reuses the variable name `missing`:
   ```python
   missing = set(threshold_names) - set(sig.parameters)
   ```
   but `missing` is already bound as a `list[str]` earlier in the same `clean_traits_for_analysis` scope (line 1007: `missing = [c for c in trait_cols if c not in df.columns]`). mypy infers the variable from its first assignment and rejects the `set[str]` rebind:
   `error: Incompatible types in assignment (expression has type "set[str]", variable has type "list[str]") [assignment]`
   Fix: rename the new variable, e.g. `missing_params = set(threshold_names) - set(sig.parameters)` (and update the `assert`/message).

2. **Black fails** — `tests/test_visualization.py`. The new `test_fallback_thresholds_are_canonical_when_keys_omitted` writes the assertions as `assert any(...), (\n    f"...")`; this repo's Black wants the message on the same line as the closing paren:
   ```python
   assert any(
       abs(y - 0.2) < 1e-9 for y in horizontal_ys
   ), f"expected canonical 0.2 NaN threshold line; got {horizontal_ys}"
   ```
   Fix: `uv run black tests/test_visualization.py` and commit. (CI: "1 file would be reformatted".)

## Notes (non-blocking)

- The new fallback test logic is sound — it correctly forces the simulate branch (`cleanup_log=None`) with a key-less `thresholds` dict and asserts the axhline lands at the canonical `0.2`, never the retired `0.3`. Tests are green on all three OSes; only Black/mypy gate it. Nice coverage of exactly the path the rest of the suite skipped.
- The name-drift `assert` is a good addition, but note `assert` is stripped under `python -O`; that's fine for an internal invariant (it's a developer guard against a future signature rename, not user-input validation).

Once CI is green I'll approve — there are no remaining correctness or design concerns from my side.

---

eberrigan (approved):
## Approving — `61679c75`

CI is green across all three OSes plus Lint and mypy, and both blocking items from the last round are fixed correctly and minimally:

- **mypy** — the shadowed `missing` is renamed to `missing_params` (line 1042) with all references updated; no more `list[str]`/`set[str]` collision against the `missing` at line 1007. Guard logic unchanged.
- **Black** — the new test's assertions reformatted to the repo's canonical `assert any(\n ...\n), f"..."` form.

No logic changed in the CI fix; the name-drift guard and the new fallback test both still do exactly what they should.

Stepping back over the whole PR: this is a tight, well-scoped alignment of the data-cleanup defaults to the canonical QC config. The refactor to a single source of truth (signature defaults, no `_QC_DEFAULTS` copy) is provably behavior-preserving for `clean_traits_for_analysis` (effective thresholds unchanged at `0.5/0.2/0.0/10`, since v0.1.0a3 already pinned them), the pipeline is unaffected (`CleanupTraitsStep` passes explicit config), and the only true new behavior — `apply_data_cleanup_filters`'s stricter bare defaults — is now documented in the CHANGELOG and guarded by `TestCanonicalDefaultDriftGuard` (config-equality + literal pin) and the new viz fallback test. All review feedback across both rounds was addressed thoughtfully.

LGTM. 🚀

---

PR #172, opened 2026-06-30
feat(#165): public outlier-removal entry point remove_outlier_samples

egao28:
## Summary

Adds the public **`remove_outlier_samples`** entry point — the quality-step follow-up to `clean_traits_for_analysis` (#164). It takes a clean (NaN-free) trait table and detects + removes outlier **samples** so PCA / UMAP / clustering can optionally run on outlier-trimmed data.

It **composes the existing public primitives** — `detect_outliers_mahalanobis` / `detect_outliers_isolation_forest` for detection and `remove_outliers_from_data` for row removal — and defines no detection/removal logic of its own, so its semantics cannot drift from the QC pipeline outlier steps (the #116/#164 single-source-of-truth pattern).

Closes #165.

## What it does

`clean_traits_for_analysis` → **(optional)** `remove_outlier_samples` → `perform_pca_analysis` / UMAP / clustering.

- **Preconditions (before any detector runs):** NaN-free trait columns (wrapped error points to `clean_traits_for_analysis`) and a unique index — both correctness guards for one-to-one index alignment (the detectors run PCA which silently `dropna()`s rows; removal is label-based).
- **Method selection:** `method="mahalanobis"` (default, chi-squared threshold) or `"isolation_forest"`; unknown method fails fast. Per-method `**detect_kwargs` (`contamination`, `chi2_percentile`, …) are forwarded unchanged.
- **Output readiness:** re-applies the #164 gates (≥2 samples, ≥1 non-constant trait). On failure raises `OutlierRemovalError` (a `ValueError`) **carrying the report** so a caller can still inspect what would have been removed.
- **Warnings:** over-removal (> 50%, before the gates) and the `p > n` regime.
- **Returns** `(trimmed_df, outlier_report)` — an auditable, JSON-serializable dict (`method`, `method_params`, `random_state`, counts, `removal_fraction`, `outlier_indices`, `outlier_barcodes`, threshold/`n_components`/`goodness_of_fit` for Mahalanobis, `None` for isolation forest).

## Changes

- **New** `src/sleap_roots_analyze/outlier_removal.py` (composition + report + readiness guards), exported from `__all__`.
- **New** `tests/test_remove_outlier_samples.py` — 37 tests (composition, preconditions, readiness, report schema + JSON adversarial, determinism, detector-failure surfacing, misuse, single-source-of-truth, public API).
- `docs/API.md`: new entry + the two previously-undocumented composed primitives (`detect_outliers_isolation_forest`, `remove_outliers_from_data`).
- `docs/CHANGELOG.md`: `[Unreleased] → Added (#165)`.
- `tests/reproducibility_cases.py` + `tests/test_reproducibility.py`: registered the new `random_state` function (auto-discovered by the package-wide determinism sweep) and updated the pinned anchors.
- OpenSpec change `add-outlier-removal-entry-point` (proposal / design / tasks / spec).

**No behavior change** to any existing function or pipeline step (additive module + one `__all__` entry).

## Testing

- `tests/test_remove_outlier_samples.py`: **37 passed**. Fixtures were tuned against the real detector so default Mahalanobis flags *exactly* the injected outliers.
- `tests/test_reproducibility.py`: **35 passed** (new case swept + coverage guard).
- Full non-integration suite green after merging `main`.
- mypy baseline filter: **new: 0**. `black --check` + `ruff check` clean on all touched files.

> `main` (including #171 — the #167 cleanup-default alignment) is merged into this branch, so the PR is current. The `[Unreleased]` changelog carries both the #165 `Added` and the #167 `Changed` entries. The new code depends only on #164 helpers, not on the #171 default changes.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -90,8 +90,8 @@ apply_data_cleanup_filters(
     df: pd.DataFrame,
     trait_cols: List[str],
     max_zeros_per_trait: float = 0.5,
-    max_nans_per_trait: float = 0.3,
-    max_nans_per_sample: float = 0.2,
+    max_nans_per_trait: float = 0.2,
+    max_nans_per_sample: float = 0.0,
     min_samples_per_trait: int = 10,
     barcode_col: str = "Barcode",
     genotype_col: str = "geno",
@@ -178,8 +178,8 @@ Validation runs in order: (1) empty input, (2) no NaN in surviving traits,
 > (`max_zeros_per_trait=0.5`, `max_nans_per_trait=0.2`, `max_nans_per_sample=0.0`,
 > `min_samples_per_trait=10`), so the analysis-ready frame matches what the QC
 > pipeline produces rather than a looser clean. (`apply_data_cleanup_filters`' own
-> signature defaults are looser — `max_nans_per_trait=0.3`, `max_nans_per_sample=0.2`;
-> aligning them is tracked in #167.) Caller kwargs override. With
+> signature defaults now equal these canonical values — `max_nans_per_trait=0.2`,
+> `max_nans_per_sample=0.0` — aligned in #167.) Caller kwargs override. With
 > `max_nans_per_sample=0.0`, any sample carrying a NaN in a surviving trait is
 > dropped. Two samples is the runnability floor only.
 
--- a/docs/API.md
+++ b/docs/API.md
@@ -1041,12 +1041,17 @@ so the trimmed frame stays analysis-ready.
 
 **Warns:**
 - `UserWarning`: when the removed fraction exceeds 0.5 (likely mis-set
-  `contamination`/threshold), and in the `p > n` regime after trimming.
+  `contamination`/threshold), and in the `p > n` regime after trimming. On the
+  Mahalanobis path it also warns on a small sample (`n < 30` — fragile chi-squared
+  tail / covariance) and when the detector's chi-squared goodness-of-fit reports the
+  distributional assumption is violated (so the `chi2_percentile` threshold's meaning
+  is questionable).
 
 **Raises:**
 - `ValueError`: empty input; duplicate columns; explicit `trait_cols`
-  missing/non-numeric; unknown `method`; non-unique index; NaN in trait columns
-  (message points to `clean_traits_for_analysis`); or a detector failure.
+  missing/non-numeric; unknown `method`; unknown / cross-method `**detect_kwargs`
+  for the chosen detector (names the keys + supported set); non-unique index; NaN in
+  trait columns (message points to `clean_traits_for_analysis`); or a detector failure.
 - `OutlierRemovalError` (a `ValueError`): when trimming leaves <2 samples or no
   non-constant trait; carries `outlier_report` as an attribute.
 
--- a/docs/API.md
+++ b/docs/API.md
@@ -979,6 +979,93 @@ Outlier detection using Mahalanobis distance on PCA-transformed data.
 
 ### Functions
 
+#### `remove_outlier_samples`
+
+```python
+remove_outlier_samples(
+    clean_df: pd.DataFrame,
+    trait_cols: Optional[List[str]] = None,
+    *,
+    method: str = "mahalanobis",
+    barcode_col: str = "Barcode",
+    genotype_col: str = "geno",
+    replicate_col: Optional[str] = "rep",
+    random_state: int = 42,
+    **detect_kwargs,
+) -> Tuple[pd.DataFrame, Dict]
+```
+
+Public outlier-removal entry point — the **quality**-step follow-up to
+[`clean_traits_for_analysis`](#clean_traits_for_analysis) (#164). Takes a clean
+(NaN-free) trait table and detects + removes outlier **samples** so PCA / UMAP /
+clustering can optionally run on outlier-trimmed data. Composes the existing
+public primitives — [`detect_outliers_mahalanobis`](#detect_outliers_mahalanobis)
+/ [`detect_outliers_isolation_forest`](#detect_outliers_isolation_forest) for
+detection and [`remove_outliers_from_data`](#remove_outliers_from_data) for row
+removal — and defines no detection/removal logic of its own, so its semantics
+cannot drift from the QC pipeline's outlier steps. Intended chain:
+`clean_traits_for_analysis` → (optional) `remove_outlier_samples` →
+`perform_pca_analysis` / UMAP / clustering.
+
+Enforces, before any detector runs, a **NaN-free** and **unique-index** input
+(both correctness guards for one-to-one index alignment — the detectors run PCA
+which silently drops NaN rows, and removal is label-based). After removal it
+re-applies the #164 readiness gates (≥2 surviving samples, ≥1 non-constant trait)
+so the trimmed frame stays analysis-ready.
+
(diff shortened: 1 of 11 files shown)


eberrigan (commented):
> **Verdict: COMMENT**

## Review Summary

A genuinely well-built entry point. It faithfully **composes the existing public primitives** (`detect_outliers_mahalanobis` / `detect_outliers_isolation_forest` / `remove_outliers_from_data`) with no algorithm duplication, closely mirrors its #164 sibling `clean_traits_for_analysis` (guard ordering, message style, `stacklevel=2`, `convert_to_json_serializable`, p>n / over-removal warnings), and is exceptionally well-tested (37 tests with a near-complete 1:1 spec-scenario→test mapping). CI is green on all three OSes + lint + mypy.

I had the review team independently verify the highest-risk question — **index alignment** — end-to-end, and it **holds**: the detectors report *label* indices (`df.dropna().index` → `identify_outliers_from_distances(indices=...)`), `remove_outliers_from_data` drops by *label*, and the NaN-free + unique-index preconditions make them consistent. Two reviewers confirmed empirically (filtered `RangeIndex` *and* `Barcode`-string index → the exact injected outlier removed; the unique-index precondition is *sufficient* and does not secretly require a 0..n-1 index). All stochastic paths (isolation forest, `robust_covariance`, randomized SVD) are seeded. No wrong-rows bug, no swallowed detector failures, input is not mutated, the report is compact + JSON-serializable. **No blocking issues.**

## Important

1. **Unknown / cross-method `**detect_kwargs` silently drops from the report *and* raises an undocumented bare `TypeError`** — `outlier_removal.py:312-326` (found independently by two reviewers; no test covers it). The detectors have fixed signatures (no `**kwargs`), so:
   - A typo'd or cross-method kwarg (e.g. `chi2_percentile` with `method="isolation_forest"`, `contamination` with the default `method="mahalanobis"`, or `chi2_percentil=99.0`) is **silently absent** from `method_params` (the reflection comprehension only iterates the detector's real parameters), so the report gives no hint the caller's intent was ignored.
   - It then raises `TypeError: detect_outliers_mahalanobis() got an unexpected keyword argument ...`, which is **not** in the `Raises:` docstring (lines 226-234, which promise only `ValueError`/`OutlierRemovalError`) nor `docs/API.md`, and leaks the internal detector name.

   This is the one place the function's otherwise-consistent "fail fast with an actionable, prefixed `ValueError`" discipline breaks down — and passing an isolation-forest-only kwarg to the default Mahalanobis method is an easy mistake. Recommend: after building `method_params`, validate `set(detect_kwargs) - set(valid_param_names)` and raise a `ValueError` naming the unknown keys + the supported set for that method (mirroring the unknown-`method` guard at lines 268-272), and add a test.

2. **The default silently removes ~2.5% of clean data, and the computed goodness-of-fit / small-n signals are never surfaced at runtime.** `method="mahalanobis", chi2_percentile=97.5` trims ~the top 2.5% *by construction* even on outlier-free data — honestly disclosed in the docstring `Note:` (lines 185-192), but the over-removal rail only fires above 50% (`_OVER_REMOVAL_GUARD`), so a routine ~2.5% trim of genuinely-clean plants passes with no signal. `goodness_of_fit` (the chi-squared KS test) is computed and threaded into the report, but the entry point never inspects `goodness_of_fit["distributional_assumption_valid"]` to warn. There is also no small-n guard (only the floor of 2 and the p>n warning), so n=5 with a meaningless chi-squared tail / near-singular covariance trims silently. Recommend emitting a `UserWarning` when `distributional_assumption_valid` is False (mirroring the p>n warning), and ideally a small-n (e.g. n<30) warning on the Mahalanobis path. This is a scientific-quality gap, not a mechanical bug — reasonable to address here or as a fast follow-up, but a biologist running defaults on a real plate should get a signal.

3. **A couple of test-tightening gaps (the suite is strong overall):**
   - `test_p_greater_than_n_warns_and_returns` uses an n=8 fixture that removes 5/8 (62.5%), so the over-removal `UserWarning` *also* fires; `pytest.warns(match="p > n")` passes but silently tolerates a second warning, so it could pass on the wrong warning if the p>n message were ever dropped. Use a fixture that triggers p>n without breaching 50%, or assert exactly the p>n warning.
   - Kwarg-forwarding *to the detector* and "detection operates over exactly `clean_df[trait_cols]`" are asserted only indirectly (via the echoed `method_params` / result counts), not via a detector spy — `method_params` would echo an override even if the kwarg were dropped before the call. A spy asserting the detector's received args would close both.
   - The exact-index Mahalanobis oracle (`[5,17,28]`) is brittle by nature (couples to detector internals), but it's correctly confined to one test while the gate/over-removal/p>n fixtures use robust `distance_threshold` cutoffs. Noting the single point of fragility, not asking to change it.

## Suggestions

- `outlier_removal.py:343-345`: `n_outliers = n_input - n_output` derives the count from the trimmed-frame delta rather than `len(outlier_indices)`. This is actually the *safer* choice (it matches what `remove_outliers_from_data` dropped after its internal `valid_outlier_indices` filter), so it's correct — add a one-line comment, or assert `n_outliers == len(outlier_indices)` to make the invariant self-checking, since the preconditions guarantee they agree.
- Near-singular covariance is silently absorbed by the `inv`→`pinv` fallback in `calculate_mahalanobis_distances` with no flag in the report; consider recording a `covariance_singular` provenance flag.
- The detector-failure guard's second clause (`"outlier_indices" not in detect_result`) is unreachable against the current detectors (which always include `"outlier_indices": []` on error) — harmless belt-and-suspenders, fine to keep.

---

eberrigan (approved):
## Approving — `4637586b`

Thanks @egao28 — verified against the commit, and this closes every item from the review (plus the suggestion). CI is green on ubuntu/windows/mac + lint + mypy + the determinism/golden gates.

**Kwarg footgun — fixed.** The new guard rejects unknown / cross-method `detect_kwargs` up front with an actionable, prefixed `ValueError` naming the bad keys and the supported set (`outlier_removal.py`), before the detector call — so a typo or an isolation-forest knob passed to the default Mahalanobis method no longer silently drops from `method_params` or leaks a bare `TypeError`. Documented in `Raises:` and covered by `test_unknown_detect_kwarg_rejected_before_detection` (contamination→mahalanobis, chi2_percentile→isolation_forest, and a typo).

**Silent ~2.5% trim — addressed, more than asked.** The Mahalanobis path now emits a small-n `UserWarning` (`_SMALL_N_MAHALANOBIS = 30`) and a goodness-of-fit warning when `distributional_assumption_valid is False`, both before the readiness gates so they're observable even on an aggressive trim. The `is False` check is precise (won't fire for isolation forest's `None`), and `test_no_quality_warnings_on_clean_large_sample` guards against spurious warnings. I'd flagged this as reasonable to defer to a follow-up — nice to see it landed here.

**Test tightening — done.** `test_p_greater_than_n_warns_without_over_removal` isolates the p>n warning, and `test_detector_receives_trait_subframe_and_forwarded_kwargs` adds the detector spy that directly asserts the exact `clean_df[trait_cols]` subframe and forwarded kwargs — closing the previously-indirect assertions. The `assert n_outliers == len(outlier_indices)` self-checking invariant is in too.

Nothing outstanding on my side. The core guarantee — correct label-based index alignment (no wrong-rows removal), fully seeded stochastic paths, faithful composition of the existing primitives, and a thorough spec→test mapping — held throughout. LGTM. 🚀

---

PR #175, opened 2026-07-01
feat(#173): plot_outlier_analysis — public outlier-plotting entry point

egao28:
## Summary

Adds **`plot_outlier_analysis`**, the public IO-free **plotting** sibling of
`remove_outlier_samples` (#165) — the immediate consumer is the bloom-mcp `remove_outliers`
tool's optional `include_plots`. Follow-up to #165, which explicitly deferred plots. Rides
`0.1.0a4` (#174) alongside #165.

Closes #173.

## What it adds

- **`plot_outlier_analysis(clean_df, trait_cols=None, *, method="mahalanobis", random_state=42, which=None, **detect_kwargs) -> Dict[str, plt.Figure]`**
  — re-detects outliers deterministically (same detector/seed/params → the same set
  `remove_outlier_samples` removes), then composes the existing public `create_*_outlier` figure
  functions. **IO-free**: returns open `Figure`s; the caller saves/persists. Covers `mahalanobis`
  + `isolation_forest`; `which` (a key or list) narrows the returned figures.
- **`_select_outlier_figures(df, results, method, which=None, genotype_col=None)`** — the shared
  per-method figure-selection, the single source of truth used by both the entry point and the
  pipeline step.

## Ratified design decisions (updated #173)

- **B1** — `create_pca_outlier_plot` (and the pca/kmeans/gmm/hierarchical plots) stay
  **pipeline-only**. The original #173 mapping fed a *Mahalanobis* result into
  `create_pca_outlier_plot`, which reads `detect_outliers_pca` reconstruction keys
  (`cumulative_variance`/`reconstruction_errors`) → `KeyError`/blank figure; the Mahalanobis
  PC-projection view is already drawn by `create_mahalanobis_outlier_plots`
  (`mahalanobis_pc_analysis`).
- **D3** — `VisualizeOutliersStep` delegates figure selection to `_select_outlier_figures` on its
  **own pre-computed** `outlier_results` (no re-detection), preserving the pipeline's configured
  detector params. Its rendered output — figures, filenames, count — is **unchanged**.

## Preconditions & safety

NaN-free + unique-index preconditions (matching `remove_outlier_samples`, so the "figures match the
trimmed table" guarantee is sound and the per-genotype indexing is aligned); detector failures
surfaced *before* delegating (the `create_*` functions silently return `{}` on error input);
unknown/cross-method `detect_kwargs` and unavailable `which` keys rejected fail-fast.

## Testing / verification

- New `tests/test_plot_outlier_analysis.py` + a step-refactor filename regression in
  `tests/test_step_visualize_outliers.py` (realistic detector output; asserts the step's mahal/IF
  figures equal the `create_*` keys and one cross-method per-genotype figure).
- **173 passed** across the affected suites; determinism-matches-removal verified for both methods.
- `mypy` baseline `new: 0`; `ruff` / `black` / public-API-docs audit clean; `openspec validate
  --strict` valid.
- Reproducibility: `plot_outlier_analysis` registered in `EXCLUDED` (returns Figures the sweep
  can't compare; adds no new stochastic step over the already-swept `detect_outliers_*`).

## Docs

New `outlier_visualization` section in `docs/API.md` (+ backfilled the composed
`create_mahalanobis_outlier_plots` / `create_isolation_forest_plots` /
`create_outliers_per_genotype_plot` entries the new function references); `docs/CHANGELOG.md`
`[Unreleased]` entry.

## OpenSpec

`add-outlier-plotting-entry-point` — capabilities **`outlier-visualization`** (new) +
**`visualization-pipeline`** (transparent step-refactor delta). Went through a 5-agent adversarial
`/review-openspec` before implementation; B1/D3 ratified in the updated #173.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -992,8 +992,9 @@ remove_outlier_samples(
     genotype_col: str = "geno",
     replicate_col: Optional[str] = "rep",
     random_state: int = 42,
+    return_detector_result: bool = False,
     **detect_kwargs,
-) -> Tuple[pd.DataFrame, Dict]
+) -> Union[Tuple[pd.DataFrame, Dict], Tuple[pd.DataFrame, Dict, Dict]]
 ```
 
 Public outlier-removal entry point — the **quality**-step follow-up to
@@ -1027,18 +1028,25 @@ so the trimmed frame stays analysis-ready.
 - `barcode_col` / `genotype_col` / `replicate_col`: metadata columns excluded from
   inferred traits (`barcode_col` is also read to populate `outlier_barcodes`)
 - `random_state`: Seed forwarded to the detector for reproducibility (default: 42)
+- `return_detector_result`: When `True`, additionally return the **raw** detector result
+  dict (a third tuple element — the per-sample arrays the compact report omits) so a
+  consumer can detect once and feed it to
+  [`select_outlier_figures`](#select_outlier_figures) without a redundant second
+  detection. Default `False` preserves the 2-tuple compact-report contract.
 - `**detect_kwargs`: Per-method parameters forwarded to the chosen detector
   (`contamination` for isolation forest; `chi2_percentile`, `variance_threshold`,
   `use_chi_squared`, `distance_threshold`, `robust_covariance` for Mahalanobis)
 
 **Returns:**
-- `Tuple[pd.DataFrame, Dict]`: `(trimmed_df, outlier_report)`. `trimmed_df` is
-  `clean_df` with the flagged rows removed (all columns preserved). `outlier_report`
-  is an auditable, JSON-serializable dict: `method`, `method_params`, `random_state`,
-  `n_input_samples`, `n_outliers`, `n_output_samples`, `removal_fraction`,
-  `outlier_indices`, `outlier_barcodes`, `threshold_type`, `threshold_value`,
-  `n_components`, `variance_threshold`, `goodness_of_fit` (the last four are
-  populated for Mahalanobis and `None` for isolation forest).
+- `(trimmed_df, outlier_report)` — or `(trimmed_df, outlier_report, detector_result)`
+  when `return_detector_result=True`. `trimmed_df` is `clean_df` with the flagged rows
+  removed (all columns preserved). `outlier_report` is an auditable, JSON-serializable
+  dict: `method`, `method_params`, `random_state`, `n_input_samples`, `n_outliers`,
+  `n_output_samples`, `removal_fraction`, `outlier_indices`, `outlier_barcodes`,
+  `threshold_type`, `threshold_value`, `n_components`, `variance_threshold`,
+  `goodness_of_fit` (the last four are populated for Mahalanobis and `None` for
+  isolation forest). `detector_result` (when returned) is the unmodified
+  `detect_outliers_*` dict.
 
 **Warns:**
 - `UserWarning`: when the removed fraction exceeds 0.5 (likely mis-set
@@ -1385,8 +1393,9 @@ plt.show()
 ## `outlier_visualization` Module
 
 Outlier-detection figure composition. The `create_*_outlier` functions each render an
-individual figure; `plot_outlier_analysis` composes the method-appropriate set for a
-consumer (e.g. the bloom-mcp `remove_outliers` tool's optional plots).
+individual figure; `select_outlier_figures` selects the method-appropriate set from a
+pre-computed detector result, and `plot_outlier_analysis` re-detects and composes them
+for a consumer (e.g. the bloom-mcp `remove_outliers` tool's optional plots).
 
 ### Functions
 
@@ -1398,6 +1407,9 @@ plot_outlier_analysis(
     trait_cols: Optional[List[str]] = None,
     *,
     method: str = "mahalanobis",
+    barcode_col: str = "Barcode",
+    genotype_col: str = "geno",
+    replicate_col: Optional[str] = "rep",
     random_state: Optional[int] = 42,
     which: Optional[Union[str, List[str]]] = None,
     **detect_kwargs,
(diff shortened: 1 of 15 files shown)


eberrigan (changes requested):
## Review Summary

This is a careful, faithful implementation of #173 — the DAG step refactor is a clean key/filename-identical extraction, the determinism-matches-removal guarantee is directly tested for both methods, CI is green on all three OSes, and the reproducibility `EXCLUDED` registration is correct. The changes below are mostly on me, not the implementation: the design I specified in #173 sets up a redundant computation (and a latent correctness gap) for the #378 `remove_outliers` consumer, and I'd rather fix it now, before the a4 public API locks, than after consumers depend on the shape. Sorry for the extra loop.

## The redundancy (and why it matters)

`plot_outlier_analysis` re-detects outliers, and #378 delegates detection to `remove_outlier_samples` first — which discards its raw detector result (it keeps only a compact report). So with `include_plots=True`, `remove_outliers` runs the detector **twice** on the same frame: once to trim, once to plot (a full PCA-SVD, or a 100-tree Isolation Forest fit, repeated). Two things I got wrong in #173 cause this:

1. I kept the precomputed-results helper (`_select_outlier_figures`) private, so a consumer that already holds a detector result has no non-redundant door.
2. I didn't give `plot_outlier_analysis` the metadata-column params that `remove_outlier_samples` has — so the "same samples as removal" guarantee can silently break (see Blocking #3).

## Blocking

**1. `which`-filter leaks figures.** `_select_outlier_figures` renders the full set, then narrows to `which` without closing the dropped figures — `plot_outlier_analysis(df, method="mahalanobis", which="mahalanobis_outlier_detection")` returns 1 figure but leaves 3 orphaned open figures in matplotlib's registry (unbounded leak in a long-running MCP server, plus wasted render). The test fixture's `plt.close("all")` masks it. Build only the requested figures, or `plt.close` the dropped ones.

**2. `random_state: int = 42` contradicts the docs + test.** `docs/API.md` says `Optional[int]` and `test_accepts_random_state_none` passes `None`. Change the annotation to `random_state: Optional[int] = 42`.

**3. Add `barcode_col` / `genotype_col` / `replicate_col` to `plot_outlier_analysis`, matching `remove_outlier_samples`.** Today it resolves trait columns with defaults only (and hardcodes `"geno"`). A caller who passes non-default metadata columns to `remove_outlier_samples` but not here gets a different trait set → a plotted outlier set that silently diverges from what was trimmed. This is a correctness bug — prioritize it over the compute redundancy. (This also removes the hardcoded-`"geno"` asymmetry.)

## Single-detection fix (do this here)

**4. Promote `_select_outlier_figures` to a public `select_outlier_figures(df, results, method, which=None, genotype_col=None)`.** This is the layer-1 helper #173 already described — just make it public and add it to `__all__`.

**5. Add an additive `return_detector_result=False` flag to `remove_outlier_samples`** so it can optionally return the raw detector dict alongside the report (default preserves the current compact-report contract). Then #378 detects **once** and feeds that result to both the trim and `select_outlier_figures` — no second fit, no re-stitching, no lost readiness-gate guardrails.

Do **not** add a `results=` param to `plot_outlier_analysis` instead — it makes `random_state`/`detect_kwargs` silently ignored when `results` is passed, and it doesn't actually fix #378 (the result is discarded before plotting anyway).

## Important

- **`random_state` inside `**detect_kwargs`** bypasses fail-fast validation: `plot_outlier_analysis(df, **{"random_state": 7})` silently overrides the seed, and passing it both ways throws a raw `TypeError` before any actionable `ValueError`. Mirrors `remove_outlier_samples`, so fix consistently if you touch that path for #5.
- **Missing edge-case tests** for guards that exist: duplicate column names (step 2), non-numeric explicit `trait_cols` (step 3), and an *all*-NaN column (only a single-NaN cell is tested today).
- **"Byte-identical" is overstated** in the step-test docstring and the two `visualize_outliers.py` comments — the tests prove figure-key/filename equivalence, not byte-equal PNGs. Reword to "same figure keys/filenames."

## Suggestions

- Document figure lifecycle in `docs/API.md`: the caller owns the returned Figures — `plt.close(fig)` those you don't retain to avoid `max_open_warning`/memory growth in a long-running server.
- Add a regression test asserting `len(plt.get_fignums())` after a `which`-narrowed call equals `len(returned)` — it fails today and would pin the #1 fix.
- Use a neutral prefix in `_select_outlier_figures`' `which`-missing error (it says `plot_outlier_analysis:` even when the pipeline step is the caller).

---

eberrigan (approved):
## Review Summary

The fix commit (`be78b35`) addresses everything from the previous review — cleanly, and with tests that assert each fix rather than just exercising the new path. The redundant-detection concern is fully resolved: a consumer can now detect once via `remove_outlier_samples(..., return_detector_result=True)` and feed the raw result to the public `select_outlier_figures`, with no second fit. CI is green on all three OSes including the mypy baseline. Approving.

## Verified fixed

- **Figure leak (was Blocking #1):** `select_outlier_figures` now closes the built-but-unrequested figures on the normal path *and* closes everything before raising on a missing `which` key — and it skips rendering the per-genotype figure when `which` doesn't ask for it (avoids the wasted render entirely). `test_which_narrow_does_not_leak_figures` asserts `len(plt.get_fignums()) == len(figs)`; `test_which_missing_key_leaves_no_open_figures` asserts `plt.get_fignums() == []` after the raise. Nicely done.
- **`random_state` annotation (was Blocking #2):** now `Optional[int] = 42`, and the seed flows through a `Dict[str, Any]` splat so a `None` seed doesn't trip the detectors' narrower `int` annotation. Good detail.
- **Metadata-column divergence (was Blocking #3):** `barcode_col`/`genotype_col`/`replicate_col` added and forwarded to `get_trait_columns`; the hardcoded `"geno"` is gone (the per-genotype figure now uses the caller's `genotype_col`). `test_metadata_columns_forwarded_to_trait_resolution` and `test_custom_genotype_col_drives_per_genotype_figure` lock this in — the "same samples as removal" guarantee now actually holds under non-default metadata columns.
- **Single-detection path (was #4/#5):** `_select_outlier_figures` → public `select_outlier_figures` (in `__all__`); `return_detector_result=False` added to `remove_outlier_samples` as an additive, backward-compatible 3-tuple. `test_return_detector_result_flag` confirms the default 2-tuple contract is preserved and the raw dict carries the per-sample arrays the compact report omits; `test_detector_result_from_removal_feeds_selection` proves the end-to-end reuse with matching indices.
- **Missing edge-case tests (was Important):** `test_duplicate_columns_rejected`, `test_non_numeric_explicit_trait_cols_rejected`, `test_all_nan_column_rejected` all added.
- **"Byte-identical" wording + figure-lifecycle docs:** step comments reworded to "same figure keys/filenames"; the `plot_outlier_analysis`/`select_outlier_figures` docstrings and API.md now document that the caller owns the returned figures.

## Suggestion (non-blocking, follow-up)

- The new return type is `Union[Tuple[df, report], Tuple[df, report, detector_result]]`. That's correct, but a downstream typed consumer (bloom-mcp) writing `trimmed, report = remove_outlier_samples(df)` will see a possible 3-tuple and can get a spurious unpack error. Adding `@overload`s keyed on `return_detector_result: Literal[False]` / `Literal[True]` would keep both call shapes precisely typed for consumers. Fine as a follow-up — this repo's own mypy baseline is green.

---

Issue #80, opened 2026-03-05
UMAP and downstream steps use unfiltered traits after PCA zero-variance filtering


eberrigan:
## Problem

Follow-up to #74. The PCA fix (#76) correctly uses `feature_names` (zero-variance traits removed) within the PCA step itself, but **does not update `trait_names` in metadata**. Downstream steps that read `trait_names` still get the original unfiltered list.

### Data flow inconsistency

```
Step 2 (Statistics): metadata["trait_names"] = ["A", "B", "C", "D"]  (4 traits, D is constant)
Step 3 (PCA):        Uses feature_names = ["A", "B", "C"]            (3 traits, D filtered)
                     But metadata["trait_names"] = ["A", "B", "C", "D"]  (still 4!)
Step 4 (UMAP):       Reads metadata["trait_names"] → operates on ALL 4 traits including D
Step 9 (Figures):    Reads metadata["trait_names"] → may plot traits not used in PCA
```

### Affected steps

- **`umap_analysis.py:81-85,96,107`** — reads `trait_names`, passes all traits (including zero-variance) to `perform_umap_analysis()`. UMAP's `StandardScaler` handles constant columns gracefully (produces zeros), so no crash, but it's wasted computation and the logged `n_traits` count is inflated.

- **`generate_static_figures.py:121`** — reads `trait_names` for trait distributions. May create histograms/boxplots for zero-variance traits that are meaningless.

- **`pca_analysis.py:166-174`** — stores `excluded_zero_variance_traits` and `n_traits_after_filtering` in metadata, but does NOT update `trait_names` to the filtered list.

## Expected Behavior

After PCA filtering, `metadata["trait_names"]` should reflect the actual traits used. The original list should be preserved under a separate key (e.g., `original_trait_names`) for traceability.

## Proposed Fix

In `pca_analysis.py`, after computing `feature_names` and `excluded_traits`:
```python
metadata = {
    **prev_result.metadata,
    "trait_names": list(feature_names),           # Update to filtered
    "original_trait_names": list(trait_cols),      # Preserve original
    ...
}
```

## Reproduction

Use the test fixture from `test_viz_pipeline_zero_variance.py` — 4 variable + 4 constant traits. After PCA, check that UMAP receives 4 traits (not 8).

Issue #74, opened 2026-02-20
Viz pipeline PCA fails when traits have zero variance (index mismatch)


eberrigan:
## Bug Description

The Viz pipeline PCA step fails with a shape mismatch error when any trait columns have zero variance (constant values).

## Error Message

```
Task 03_pca_analysis failed: Shape of passed values is (263, 4), indices imply (267, 4)
```

## Root Cause

**Location**: `src/sleap_roots_analyze/pipeline/steps/pca_analysis.py` lines 115-120

```python
loadings_df = pd.DataFrame(
    pca_results["loadings"],
    index=trait_cols,  # Original trait columns (e.g., 267)
    columns=[f"PC{i + 1}" for i in range(n_components)],
)
```

The step uses `trait_cols` (267 traits) as the index, but `perform_pca_analysis()` internally drops zero-variance columns.

**Internal filtering**: `src/sleap_roots_analyze/pca.py` lines 641-645 in `standardize_data()`:

```python
variances = df_clean.var(ddof=0)
non_zero_var_cols = variances[variances > 0].index
df_clean = df_clean[non_zero_var_cols]
```

## Reproduction

```python
# Dataset: Alfalfa GWAS Wave 1 Day 0
# Zero-variance traits (all day-0 seedlings have exactly 1 lateral root):
#   - Lateral Count Med: all values = 1.0
#   - Lateral Count Min: all values = 1
#   - Lateral Count P25: all values = 1.0
#   - Lateral Count P5: all values = 1.0
```

## Similar Issues to Check

Search for other places where an index/column list might mismatch after internal filtering:

```bash
# Find all DataFrame creations with explicit index
grep -rn "pd.DataFrame(" src/sleap_roots_analyze/pipeline/steps/ | grep -i index
```

Potential issues found:
- `pca_analysis.py:115` - **THIS BUG** (loadings_df uses trait_cols)
- `pca_analysis.py:107` - pc_scores_df uses data_clean.index (CORRECT)
- `umap_analysis.py:130` - embedding_df uses data_clean.index (CORRECT)

### 1. Write failing tests first

Add to `tests/test_step_pca_analysis.py`:

```python
class TestPCAZeroVariance:
    """Test PCA handling of zero-variance traits."""

    def test_pca_with_zero_variance_traits(self, config, prev_result, tmp_path):
        """Test PCA handles traits with zero variance (constant values)."""
        data = pd.DataFrame({
[...]

PR #178, opened 2026-07-08
feat(cleanup): drop + name zero-variance (constant) traits at cleaning time (#177)

eberrigan:
## What this PR does

Implements the configurable **zero-variance / constant-trait filter** in the cleanup step (design doc: `docs/superpowers/specs/2026-07-08-zero-variance-trait-cleanup-filter-design.md`). A constant trait is now **dropped and named at cleaning time** instead of being silently dropped later by PCA. Tracking issue: #177.

> Updated from the original design-only handoff — this branch now carries the full implementation, an OpenSpec change, and tests.

## Why

A constant (zero-variance) trait carries no information for multivariate analysis, but today it's handled **silently and late**: `standardize_data` drops it inside PCA without naming it, cleanup has no variance filter, and `clean_traits_for_analysis` only validates that *some* non-constant trait survives (never dropping the individual constants). So a constant trait passes cleaning into the "analysis-ready" set and is then silently dropped by PCA — or turns into a hard error on the direct `clean_traits_for_analysis` + `perform_pca_analysis` path used by bloom-mcp's `pca_analysis` tool. The honest, systemic fix is to drop + name constants **at cleaning time** so every consumer benefits.

## What changed

- New internal `remove_zero_variance_traits(df, trait_cols, min_variance=0.0)` — drops `var(ddof=0) <= min_variance` (exact complement of `standardize_data`'s `> 0` keep test). Skips non-numeric columns; not exported.
- Wired as the **final step** of `apply_data_cleanup_filters`, evaluated on the **post-sample-removal** frame (a trait can go constant only after NaN rows are dropped).
- **Re-checked** inside `clean_traits_for_analysis` after its own `dropna`, so the analysis-ready frame is constant-free even when `max_nans_per_sample` is loosened; the re-check is skipped when that dropna removes no rows.
- Config: `CleanupConfig.min_variance: float = 0.0` — threshold-only, always-on (negative disables), forwarded by `CleanupTraitsStep`; canonical-default drift guard (#167) updated; commented example added to the QC config templates.
- Log: `cleanup_log["removed_traits"]` entry with `reason="zero_variance"` (+ `variance`, `threshold`) and a `cleanup_steps` entry — **named, not silent**. `final_samples`/`final_traits`/retained-fractions stay self-consistent on the loosened-NaN path.
- Warns when `min_variance` is non-finite (NaN silently disables; `+inf` drops all) or `> 0` (raw, scale-dependent variance).
- `standardize_data` left unchanged as a belt-and-suspenders no-op.
- OpenSpec change: `openspec/changes/add-zero-variance-cleanup-filter/`.

## Behavior change (default-on)

For any dataset with a genuinely-constant trait, that column is now **absent** from cleaned output and **named** in the cleanup log. **PCA / UMAP / clustering results are unchanged** (those paths already dropped constants before fitting), and the QC PCA step's `excluded_zero_variance_traits` becomes empty when fed a cleaned frame. A degenerate `heritability=0.0` / ANOVA `p=NaN` row for a constant-nonzero trait no longer appears (`n_traits_analyzed` drops accordingly). Set `min_variance` negative to keep the old behavior. See `docs/CHANGELOG.md`.

## Tests

A dedicated `tests/test_zero_variance_cleanup.py` (23 tests) covers the granular filter, orchestrator wiring, entry-point re-check, config + step forwarding, log consistency, and the misuse warnings; the drift-guard and entry-point tests are updated. Addresses two rounds of review (mypy-baseline paydown, log self-consistency, non-numeric dtype guard, real-pipeline viz label, sample-count refresh, config templates).

## Incorporation into bloom-mcp (separate repo: [Salk-Harnessing-Plants-Initiative/bloom](https://github.com/Salk-Harnessing-Plants-Initiative/bloom))

After this lands upstream + a version bump:

1. Bump the `sleap-roots-analyze>=` pin to the release containing this filter.
2. Surface `reason="zero_variance"` in `qc_clean` — [bloom#338](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/issues/338), [bloom#356](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/issues/356).
3. Downgrade or keep the now-unreachable `pca_analysis` guard — [bloom#308](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/issues/308), [bloom#412](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/issues/412).
4. Update bloom tests/fixtures that asserted the old hard-error behavior.
5. Close [bloom#412](https://github.com/Salk-Harnessing-Plants-Initiative/bloom/issues/412) as resolved-upstream.

## Related issues (this repo)

- #80 — related bug (downstream `trait_names` after PCA zero-variance filtering); **preempted** by this change (no constant reaches PCA in the QC pipeline), but not fully fixed, so it stays valid as defense-in-depth.
- #74, #167, #168 — prior art / cleanup-default lineage.


eberrigan (commented):
> **Verdict: REQUEST_CHANGES** (posted as a comment — GitHub does not allow requesting changes on your own PR)

## Review Summary

The implementation faithfully follows the design doc and is genuinely good work: a clean granular filter mirroring the sibling contract, the two-drop-site approach, a full OpenSpec proposal, a thorough CHANGELOG, and 17 well-targeted tests (the `ddof=0` and `<=`-boundary tests actually *discriminate* — they'd fail under `ddof=1` / `<`). The core statistics are correct: the `var(ddof=0) <= 0` drop is an exact complement of `standardize_data`'s `var(ddof=0) > 0` keep, and I verified the constant-trait leak this closes really does exist on `main`.

Two blocking items: a **red required CI check** the PR doesn't address, and an **undocumented change to heritability/ANOVA outputs**. Everything else is minor.

---

## Blocking Issues

### B1. `Type check (mypy baseline)` is red — the PR never touches `.mypy-baseline.txt`

CI reports exactly 2 new violations:
```
src/sleap_roots_analyze/data_cleanup.py:882: error: "object" has no attribute "append"  [attr-defined]
src/sleap_roots_analyze/data_cleanup.py:884: error: "object" has no attribute "append"  [attr-defined]
```

**Root cause:** inside `apply_data_cleanup_filters`, `cleanup_log` is a bare dict literal (`data_cleanup.py:782`) whose values mix `int` and `list`, so mypy infers `dict[str, object]` and every `.append` on a value errors. Steps 1/2/4 hit this too — they're just already frozen in `.mypy-baseline.txt`. Step 5 adds 2 more. (The re-check appends in `clean_traits_for_analysis` don't error because there `cleanup_log` comes back typed as a bare `Dict` = `dict[Any, Any]` from the `-> Tuple[pd.DataFrame, Dict]` return annotation at `data_cleanup.py:726`.)

**Fix:** annotate `cleanup_log: Dict[str, Any]` at `data_cleanup.py:782` (`Any` is already imported) **and regenerate the baseline**:
```
uv run mypy src/sleap_roots_analyze | uv run mypy-baseline sync
```
The regen is required, not optional: annotating removes 12 recorded baseline entries (the 9 `.append` errors plus the `object`-operand errors they fed), and `mypy-baseline filter` also fails on a *stale* baseline whose recorded errors no longer reproduce (see `ci.yml:62-63`). This pays down 12 debt entries rather than adding 2 — the repo's stated ratchet philosophy.

### B2. Heritability / ANOVA outputs silently lose constant-trait rows — undocumented

A **constant-but-nonzero** trait (e.g. always `3.0`; note an all-zeros trait is already caught by the zero-inflation filter) previously survived cleanup and flowed into `StatisticalAnalysisStep` via `valid_trait_names` → `calculate_heritability_estimates` (`statistical_analysis.py:162`) and `perform_anova_by_genotype` (`:97`). It produced real reported entries: heritability `{'heritability': 0.0, 'model_type': 'no_variance', ...}` (`statistics.py:326`) and ANOVA `{'f_statistic': nan, 'p_value': nan, ...}`.

This PR drops that column at cleanup Step 5, so those rows now **vanish** from the heritability/ANOVA tables and `n_traits_analyzed` (`statistical_analysis.py:283`) changes.

The CHANGELOG explicitly reassures that "PCA / UMAP / clustering results are unchanged" — which I verified is accurate — but says **nothing about statistics**, and neither does the OpenSpec proposal. For a library whose outputs back publications, a change in heritability/ANOVA table row counts must be stated, not inferred.

To be clear: the vanished entries are statistically degenerate (H²=0 / `no_variance`, p=NaN), so this is arguably a **correctness improvement**. It just has to be documented. Fix: one "Behavior note" line in `docs/CHANGELOG.md` naming the heritability/ANOVA tables, plus ideally a regression test pinning the new row set.

---

## Important Issues

### I1. `traits_retained_fraction` goes stale when the entry-point re-check fires (newly introduced)

The re-check updates `final_traits` but not the fraction:
```python
# data_cleanup.py:1186-1187
        # Keep the orchestrator-set summary consistent with the re-check's extra drop.
        cleanup_log["final_traits"] = len(surviving)
```
`traits_retained_fraction` was set once by the orchestrator (`data_cleanup.py:900-904`). Reproduced on the loosened path (`max_nans_per_sample=0.5`): `original_traits: 3`, `final_traits: 2`, `traits_retained_fraction: 1.0` — the log **contradicts itself**. This design explicitly targets bloom-mcp reading this log programmatically, so it matters. `test_recheck_after_entry_point_dropna` (`test_zero_variance_cleanup.py:199`) asserts trait absence but never the summary fields, so it ships untested.

**Fix:** recompute `traits_retained_fraction` alongside `final_traits`, and add an assertion.

> Note: `final_samples` / `samples_retained_fraction` are *also* stale after the entry point's `dropna` — but I verified that is **pre-existing on `main`**, not introduced here. Worth fixing in the same pass (the new code makes the asymmetry glaring) but it's not a regression.

### I2. Non-numeric trait column now raises `TypeError` in the public `apply_data_cleanup_filters`

`remove_zero_variance_traits` calls `df[trait].var(ddof=0)` with no dtype guard (`data_cleanup.py:697`). The four sibling filters tolerate object dtype (`== 0`, `.isna()`, `.notna()`), so `main` passed such columns straight through. Verified A/B:
- PR head: `TypeError: could not convert string to float: 'x'`
- `origin/main`: no error, `final_traits = 2`

**Reachability is narrow** (I initially over-stated this): `get_trait_columns` *does* numeric-filter (`data_cleanup.py:163-169`), and `clean_traits_for_analysis` validates numeric, so the QC pipeline and the public entry point are both safe. It only bites direct callers of the **exported** `apply_data_cleanup_filters` passing a hand-built `trait_cols`. Still, the new docstring claims it "Mirrors the contract of the sibling trait filters" — on dtype tolerance, it does not.

**Fix:** guard with `pd.api.types.is_numeric_dtype(df[trait])` (skip, don't flag), + a test. Or document the numeric precondition.

### I3. `01_removed_traits_detail.csv` schema — empty-case fallback not updated

`cleanup_traits.py:131-142` builds the CSV from `pd.DataFrame(cleanup_log["removed_traits"])` (columns = union of whichever filters fired) but falls back to a hard-coded `columns=["trait","reason","zero_fraction","nan_fraction","n_samples"]` when empty. Zero-variance entries add `variance`, so the non-empty header is now `['trait','reason','zero_fraction','threshold','variance']` while the empty header has neither `threshold` nor `variance`. The union-instability predates this PR, but this widens it. No `src` consumer reads the file, so nothing internal breaks — external consumers see an inconsistent header.

**Fix:** add `threshold` and `variance` to the fallback list; document the union schema.

### I4. Redundant variance scan on the default path

On the canonical `max_nans_per_sample=0.0`, the entry point's `dropna` (`data_cleanup.py:1166`) is provably a no-op (Step 3 already removed every row with any NaN in a superset of `surviving`), so the re-check at `:1173` is a pure duplicate full-frame scan. Measured ~32 ms on a 5000×300 frame (`.copy()` ~10 ms + 300-column `.var()` loop ~22 ms). This takes `clean_traits_for_analysis` from 1 variance pass on `main` to 3.

**Fix** — the row-count delta is free:
```python
n_before = len(clean_df)
if surviving:
    clean_df = clean_df.dropna(subset=surviving)
if surviving and len(clean_df) != n_before:
    clean_df, surviving, zero_variance_recheck = remove_zero_variance_traits(...)
```
Preserves the loosened-NaN correctness that `test_recheck_after_entry_point_dropna` covers.

### I5. `clean_traits_for_analysis` docstring not updated for `min_variance`

Both the Notes canonical-defaults list (`data_cleanup.py:1032-1034`) and the `**cleanup_kwargs` forwarded-params list (`data_cleanup.py:1057-1059`) still enumerate only the original four thresholds, though `min_variance` is now forwarded and recorded in `effective_thresholds`.

### I6. Test gaps

- No test for a non-numeric trait column (the I2 crash path).
- No test asserting `traits_retained_fraction` / `final_traits` consistency after the re-check (I1).
- No all-NaN trait column case (`var` → NaN → not flagged; distinct from the empty-frame path).
- `CleanupConfig(min_variance=-1.0)` is never driven through `CleanupTraitsStep` — only the default-forwarding case is integration-tested.

---

## Suggestions

- **`min_variance` is scale-dependent.** It compares raw, pre-standardization variance (squared trait units), so a single global non-zero threshold mixes mm² with dimensionless ratios. Harmless at the `0.0` default (exactly "is constant"), but the docstrings (`data_cleanup.py:757-760`, `components.py:68-72`) should warn against non-zero values.
- **`min_variance` is unvalidated for NaN/inf.** `nan` silently disables the filter (`var <= nan` is always False — an undocumented second disable path); `inf` drops **every** trait. Consider a finiteness check. (Other cleanup thresholds also lack range validation, so this is consistent, not a regression.)
- **`visualization.py:856-869`** has no `elif` branch for `"zero_variance"`, so the raw string is displayed rather than a formatted label like the siblings.
- **Factor the duplicated re-check block.** `data_cleanup.py:1173-1187` duplicates Step 5's call+log-append (`:876-890`); a small `_apply_and_log_zero_variance(...)` helper would remove the duplication and give I1 a single place to be fixed correctly.
- **`01_data_traits_cleaned.csv`** now omits traits that only went constant *after* sample removal, though that file is documented as "trait removal only, before sample removal". Worth a doc note — but this exactly mirrors existing Step 4 (`remove_low_sample_traits`) behavior on `main`, so it is not a new class of bug.

---

## Verified sound — don't chase these

- **`ddof=0` complement is exact.** Cleanup drops `var(ddof=0) <= 0`; `standardize_data` keeps `> 0`. Denormals (`1e-300`) underflow to `0.0` at both sites consistently.
- **No false-positive drops.** `Series.var()` is `skipna=True`, and the entry point's `dropna` only *removes* rows — a subset of constant values stays constant. Step 5 can never flag a trait that would have varied post-`dropna`.
- **PCA / UMAP / clustering embeddings genuinely unchanged.** UMAP does *not* drop constants, but a constant column becomes exactly `0.0` after `StandardScaler` and contributes zero to Euclidean distance (verified: max pairwise distance delta = 0.0). The CHANGELOG claim is accurate.
- **`excluded_zero_variance_traits` becomes an empty list** and has no `src` consumer — benign, as designed.
- **Purity / idempotency:** `remove_zero_variance_traits` does not mutate its input; a second run removes nothing.
- **Edge paths correct:** single-row → check (3) `only 1 sample(s) remain`; multi-row all-constant → check (4) `no non-constant numeric trait remains`. Empty frame flags nothing; absent trait stays in `remaining_trait_cols`.
- **Cross-platform / determinism clean.** No new `filterwarnings`/`np.errstate`/global state; `np.float64(nan) <= 0.0` → `False` with no warning; suite passes under `-W error::RuntimeWarning`.
- **The `test_cleanup_path_reduces_to_below_two_samples` rewrite is a legitimate adaptation, not a weakening** — it passes loosening kwargs that override the defaults, and the `>=2 samples` gate still fires for the reason its docstring claims.
- **Drift guard correctly updated** in both the config-derived and pinned-literal dicts.

## Pre-existing — not this PR

- `final_samples` / `samples_retained_fraction` staleness after the entry-point `dropna` (present on `main`).
- Duplicate `get_numeric_traits_only` definition at `data_cleanup.py:172` and `:395` (present on `main`) — worth its own issue.
- `01_removed_traits_detail.csv` union-schema instability.

## TDD note

`tasks.md` is structured test-first (red → green → verify) with a clean 1:1 task↔test mapping, and the test quality is consistent with genuine test-first design. But `6ed396e` squashes implementation + all 17 tests into one commit, so there's no red commit demonstrating the tests failing first. Consistent with the repo's squash workflow — flagging only that the red phase isn't verifiable from history.

---

egao28 (commented):
> **✅ Update — all tasks addressed in `3df364e` (pushed).** Targeted suites green (23 + 132 + 183); black / ruff / mypy-baseline clean. PR title + description updated to reflect the implementation.

| Task | Status |
|------|--------|
| **1** 🔴 viz label never renders in real output | ✅ added the `zero_variance` elif to the **actual-removed-traits** branch of `create_trait_eda_plots` (not just the preview branch) |
| **2** 🔴 wrong sample count on loosened-NaN path | ✅ `clean_traits_for_analysis` now refreshes `final_samples` / `samples_retained_fraction` after its own `dropna`; test asserts `19` / `0.95` and log↔`validation_summary` agreement |
| **3** all-NaN overclaim | ✅ reworded the `remove_zero_variance_traits` docstring — the "won't be dropped by PCA" guarantee is scoped to **non-NaN** variance |
| **4** min_variance in templates | ✅ commented `min_variance` line added under `cleanup:` in the four QC templates |
| **5** regression guards | ✅ assert the empty-case CSV header includes `threshold`/`variance`, and that the canonical path logs exactly **one** zero-variance `cleanup_steps` entry (I4 skip is now observable) |
| **6** warn on bad min_variance | ✅ `apply_data_cleanup_filters` warns on non-finite (`NaN` disables, `+inf` drops all) or `> 0` (scale-dependent) `min_variance`; +2 tests |
| **7** housekeeping | ✅ cleanup step 5 added to `CleanupTraitsStep` docstring; **PR title + description updated** |

*Task 3 note:* the CHANGELOG's "constant-free" wording is accurate for the entry-point output (all-NaN traits are removed by the NaN filters before the variance step), so only the docstring carried the literal all-NaN overclaim — now fixed.

---

<sub>Original review below.</sub>

> **Verdict: COMMENT** — no blocking issues. The core is correct, well-tested, and the prior review round (B1/B2/I1–I6) is fully addressed. Below is a prioritized, actionable task list to pick up before merge.

## What's solid (verified, not assumed)

- **The central correctness claim holds.** `remove_zero_variance_traits` drops on `var(ddof=0) <= 0` (default) and `standardize_data` keeps on `var(ddof=0) > 0` (`pca.py:642-645`) — exact logical complements for every non-NaN variance, no float/epsilon gap. So a survivor is never silently re-dropped by PCA on the cleanup path.
- **Tests pass and map 1:1 to the spec.** All 21 tests in `tests/test_zero_variance_cleanup.py` plus 125 existing tests in the 4 most-affected files run green. Every OpenSpec scenario (both delta specs) has a dedicated test; assertions are precise. CI-safe on all 3 OSes.
- **All prior-review fixes are present and correct:** `cleanup_log: Dict[str, Any]` + mypy baseline paydown, non-numeric skip, empty-CSV header, guarded re-check, docstrings.
- **The statistics change is a genuine correctness improvement, documented accurately** in the CHANGELOG (degenerate `heritability=0.0` / ANOVA `p=NaN` rows for constant-nonzero traits now vanish).
- Deterministic (no RNG), stateless, no path/encoding/global-state/cross-platform concerns. Cleanly mirrors the sibling filters; DAG wiring is complete.

---

## 🔴 Do these first (small, worth landing before merge)

### Task 1 — The zero-variance plot label never renders in real pipeline output
`src/sleap_roots_analyze/visualization.py:781-794` (`create_trait_eda_plots`)

The nice `"Zero variance (var=…)"` label was added only to the **preview** branch (`if not cleanup_log:`, ~line 868), which runs only when there is **no** cleanup log. In the actual QC pipeline a populated `cleanup_log` **is** passed in, so the *other* branch runs — and it has no `zero_variance` case, so the plot shows the raw string `"zero_variance"`.

**Fix:** add the same `elif` to the actual-removed-traits branch, right before the `else:` at line 793:

```python
elif reason == "zero_variance":
    removal_reasons_dict[trait_name] = (
        f"Zero variance (var={trait_info.get('variance', 0):.2g})"
    )
```

*Check:* run a pipeline (or call `create_trait_eda_plots` with a real `cleanup_log` containing a `zero_variance` removal) and confirm the legend reads "Zero variance (var=…)".

### Task 2 — The cleanup log reports the wrong sample count on the loosened-NaN path
`src/sleap_roots_analyze/data_cleanup.py:1196-1217` (`clean_traits_for_analysis`, the re-check block)

The re-check correctly fixes `final_traits` and `traits_retained_fraction`, and the comment says it keeps the summary fields consistent "so programmatic consumers never read a self-contradictory log." But `final_samples` / `samples_retained_fraction` are set earlier (line ~910, inside `apply_data_cleanup_filters`) **before** the entry point's own `dropna` at line 1186 — and never updated. So when that `dropna` drops a row, the log says e.g. `final_samples = 20` while `validation_summary["n_samples"] = 19`.

**Fix:** after the re-check block (after line 1217), refresh the sample fields from the actual frame — unconditionally, since the count is stale whenever the `dropna` removed rows, even if no trait went constant:

```python
# The entry point's own dropna above can drop rows after apply_data_cleanup_filters
# set the sample-count summary; refresh so final_samples matches the returned frame.
cleanup_log["final_samples"] = len(clean_df)
cleanup_log["samples_retained_fraction"] = (
    len(clean_df) / cleanup_log["original_samples"]
    if cleanup_log["original_samples"] > 0
    else 0
)
```

*Add a test:* extend `test_recheck_after_entry_point_dropna` to also assert `log["final_samples"] == 19` and `log["samples_retained_fraction"] == 0.95` (it currently only checks the trait fields).

---

## 🟡 Nice to have (optional polish)

### Task 3 — Tighten an overclaim about all-NaN traits
`src/sleap_roots_analyze/data_cleanup.py:660-662` + the CHANGELOG entry

"A trait that survives cleanup will not be silently dropped later by PCA" isn't true for an **all-NaN** column: this filter keeps it (`NaN <= 0` is False) but `standardize_data` drops it (`NaN > 0` is False). Very unlikely in practice (all-NaN traits are removed earlier by the NaN/low-sample filters), so just reword to "…a trait with **non-NaN** variance that survives cleanup…".

### Task 4 — Add `min_variance` to the config templates
`configs/templates/*.yaml`

The four sibling cleanup fields (`max_nan_fraction`, `max_zeros_per_trait`, `max_nans_per_trait`, `min_samples_per_trait`) appear in the templates but `min_variance` doesn't. Add a commented line under the `cleanup:` section for discoverability.

### Task 5 — Add regression guards for two things that are fixed but untested
- The empty-case CSV header fix at `cleanup_traits.py:138-148` is *run* by `test_cleanup_step_min_variance_negative_disables` but never *asserted* — add a `pd.read_csv(...)` and check the header includes `threshold`/`variance`.
- The re-check-skip optimization (the `if surviving and len(clean_df) != n_rows_before_dropna:` guard) isn't observable in any test — assert that on the default path there is exactly **one** `"remove_zero_variance_traits"` entry in `cleanup_log["cleanup_steps"]`, not two.

### Task 6 — (Optional) Warn on a non-zero / non-finite `min_variance`
`min_variance` compares *raw, squared-unit* (scale-dependent) variance, so consider a `warnings.warn(...)` when `min_variance > 0` so users don't assume standardized-scale semantics. A `NaN` threshold silently disables the filter — a small `math.isfinite(min_variance)` guard would catch that footgun.

### Task 7 — Housekeeping
- Update the PR title/description — it still says "documentation only, no implementation code," but it now carries the full feature + 21 tests.
- `CleanupTraitsStep`'s class docstring lists cleanup steps 1–4; add step 5.

---

**Not a bug — no action needed:** the per-column `.var()` loop, the `.copy()`, and the pre/post-sample-removal diagnostic-file behavior all match how the existing sibling filters already work, so they're consistent by design.

---

egao28 (approved):
## Approve — reviewed against #177, nothing blocking (head `3df364e`)

Re-reviewed the full diff at the current head `3df364e` (not just the earlier `98e7f54`).

**All #177 requirements met:**
- `remove_zero_variance_traits(df, trait_cols, min_variance=0.0)` drops `var(ddof=0) <= min_variance`, matching `standardize_data`'s ddof=0 keep-test — so a survivor is not silently re-dropped by PCA.
- Wired as the **final** step of `apply_data_cleanup_filters` (post-sample-removal), correct because a trait can only go constant after NaN rows are dropped.
- Re-checked in `clean_traits_for_analysis` after its own dropna, gated to only re-run when the dropna actually removed rows.
- `CleanupConfig.min_variance: float = 0.0` added and forwarded through `CleanupTraitsStep`.
- Removals logged with `reason="zero_variance"` + variance + threshold; `standardize_data` untouched.

**Second-review items (all confirmed fixed in `3df364e`):**
- Real-pipeline viz label added to `removal_reasons_dict` (the populated-`cleanup_log` branch), not just the preview branch.
- `final_samples` / `samples_retained_fraction` now refreshed unconditionally after the entry-point dropna, so the log agrees with `validation_summary["n_samples"]` on the loosened-NaN path (test asserts `19` / `0.95`).
- `min_variance` warns on non-finite (NaN disables / +inf drops all) and on `> 0` (scale-dependent, squared trait units).
- Docstring all-NaN overclaim scoped to "non-NaN variance"; `min_variance` added to the 4 QC config templates; `CleanupTraitsStep` docstring lists step 5.

**Verification on `3df364e` (local, mirrors CI):**
- Targeted suites: 132 passed (`test_zero_variance_cleanup` + `test_clean_traits_entry_point` + `test_data_cleanup`).
- `ruff check src/sleap_roots_analyze` clean; `black --check` clean; `mypy | mypy-baseline filter` → `new: 0`.
- GitHub CI: 10/10 green across Ubuntu/macOS/Windows.

Edge cases are correct and well-covered (empty/all-NaN → not flagged; single-row → constant; absent/non-numeric columns skipped). Clean implementation, thorough tests. LGTM.

PR #176, opened 2026-07-02
Release v0.1.0a4

egao28:
## Release v0.1.0a4

Cuts `0.1.0a4` from current `main`. Implements #174. Bundles the outlier-removal + outlier-plotting entry points and the canonical cleanup-defaults fix — all a4 items are already merged on `main` (#172, #171, #175), so this is the issue's **recommended** bundle (plots included, not deferred to a5).

### Diff vs `main` (minimal)
- `pyproject.toml` + `uv.lock`: `0.1.0a3` → `0.1.0a4` (`__init__` is dynamic via `importlib.metadata`).
- `docs/CHANGELOG.md`: new `## [0.1.0a4] - 2026-07-02 (Pre-release)` section (moved from `[Unreleased]`, which is now empty).

### What's in this cut (already on `main`)
- **`remove_outlier_samples`** — public outlier-removal entry point (#165, PR #172).
- **`plot_outlier_analysis`** + `select_outlier_figures` + additive `remove_outlier_samples(return_detector_result=...)` — public outlier-plotting entry point (#173, PR #175).
- **Canonical cleanup-defaults fix** — `apply_data_cleanup_filters` bare defaults aligned to `CleanupConfig()` (#167, PR #171).
- (chore: OpenSpec archive — PR #170.)

### Pre-release validation (local)
- [x] `uv build` → wheel + sdist built as `0.1.0a4`
- [x] Isolated wheel smoke test: `__version__ == 0.1.0a4`; `remove_outlier_samples` / `plot_outlier_analysis` / `select_outlier_figures` import; CLI `--help` works
- [x] `tests/test_public_api.py`: 62 passed (release-safe changelog test green)
- [ ] Full matrix test suite — verified by this PR's CI

### Post-merge
1. Create GitHub release `v0.1.0a4` (`--prerelease`) → triggers `build.yml` trusted-publish to PyPI.
2. Verify `0.1.0a4` on PyPI; clean-env install + import check.

### Unblocks
- bloom-mcp pin bump to `sleap-roots-analyze>=0.1.0a4` + re-lock (mirror #327) → the `remove_outliers` tool (bloom#378) with optional `include_plots`, and dropping the `qc_clean` `_QC_DEFAULTS` temp override.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #181, opened 2026-07-09
feat(#180): UMAPResult serializable dataclass + from_umap_dict adapter

egao28:
## Why

Closes #180. The result-types epic (#130) added serializable dataclasses for PCA
(#127), heritability (#128), and clustering (#129), but **UMAP was skipped**.
`perform_umap_analysis` returned a dict with non-serializable `reducer`/`scaler` and no
`feature_names`/`n_components`/`random_state`, so a consumer needing a serializable,
self-describing result — the bloom-mcp `umap_analysis` tool
(Salk-Harnessing-Plants-Initiative/bloom#425) — had nothing to build one from.

## What

- **`UMAPResult`** — frozen JSON-serializable dataclass (sibling to `PCAResult`) in
  `result_types.py`: `embedding`, `n_neighbors`, `min_dist`, `n_components`,
  `feature_names`, `n_samples`, `standardized`, `random_state`. Excludes the fitted
  `reducer`/`scaler`; provides `to_dict()` / `to_json()` (strict `allow_nan=False`).
- **`UMAPResult.from_umap_dict(d, *, random_state=None)`** — non-mutating adapter that
  derives `n_components`/`n_samples` from the embedding shape, `standardized` from the
  scaler, and resolves the seed from the argument or the echoed dict key.
- **`perform_umap_analysis`** additionally returns `feature_names` + `random_state`
  (additive, non-breaking — `embedding`/`reducer`/`scaler`/`n_neighbors`/`min_dist`
  preserved).
- Root-export `UMAPResult`; docs (types table + `docs/CHANGELOG.md` `[Unreleased]`).

## Design notes

- `n_samples`/`n_components` are **stored fields** (not properties) — matches the #180
  field list and the pipeline's existing `n_samples` payload, so a JSON consumer gets
  them after the result crosses the boundary as a plain dict.
- `random_state` is **echoed by the producer and fallback-resolved by the adapter**, so
  the result records UMAP's actual load-bearing seed rather than one stamped on trust.
  Typed `Optional[int]` for parity with `PCAResult`.
- The proposal was vetted by a 5-agent OpenSpec review before implementation.

## Testing

- New `tests/test_umap_result.py` — **21 tests**: native-type pre-serialization
  (guards the `np.float64`-subclass laundering trap), JSON round-trip, no-sklearn-object,
  materialized `n_samples`/`n_components`, non-finite `to_json` rejection, adapter field
  mapping, `n_components == 1` nested shape, `standardized == False`, seed resolution
  (arg / echoed dict / `None`), non-mutation, same-seed determinism, producer enrichment,
  `n_neighbors` clamping, and public export.
- Full suite: **2433 passed, 31 skipped, 0 failed**. Black + Ruff clean.

## Scope

- **No version bump** — `0.1.0a5` is cut in a separate `chore` release PR (per #176),
  bundling this with #179.
- Hierarchical-clustering result work (#179) is a separate PR.


eberrigan (approved):
## Review Summary

Clean, well-scoped addition. `UMAPResult` faithfully mirrors the `PCAResult` sibling convention, the enrichment of `perform_umap_analysis` is genuinely additive/non-breaking (verified against every consumer: pipeline step, interactive visualization, reproducibility sweep, golden-embedding recompute), and the seed-resolution logic (`explicit arg wins, else echoed dict, else None`) is a real reproducibility improvement over the stamp-on-trust pattern used by earlier siblings — reasoned through carefully in `design.md`. All 21 new tests pass, CI is green on all three OSes, and every OpenSpec scenario has a corresponding test. No blocking issues from any of the five review passes (code quality, testing/TDD, statistical rigor, performance/cross-platform, behavioural correctness).

## Blocking Issues

None.

## Important Issues

- **`random_state=0` is untested.** `result_types.py`'s `from_umap_dict` correctly uses `is not None` (not truthiness) to resolve the seed, so an explicit `0` or a dict echoing `0` is preserved correctly — verified empirically. But no test pins this down. This is exactly the kind of falsy-seed footgun a future refactor to the idiomatic-but-wrong `random_state or d.get(...)` would silently break, and the design doc itself stresses that UMAP's seed is load-bearing. Suggest adding a `0`-seed sibling to `test_random_state_explicit_arg_wins` / `test_random_state_falls_back_to_dict` before merge.

## Suggestions

- `from_umap_dict`'s `embedding.shape[1]` raises a bare `IndexError` on an empty or 1-D embedding. This is within the documented contract (spec explicitly marks partial-dict behavior "unspecified," and the real producer never returns anything but 2-D), so not required — but a `ValueError` with a clearer message would be a friendlier edge for a hand-built dict.
- `standardized=False` is only reachable via a hand-built `scaler=None` dict since the real producer always standardizes — already called out in `design.md`'s Risks section and guarded by a dedicated test, just flagging for awareness.
- `UMAPResult.random_state: Optional[int]` permits a seedless-but-authoritative-looking JSON record (`random_state: null`) for a stochastic, seed-critical algorithm. The `PCAResult`-parity rationale in `design.md` is reasonable, but consider a one-line docstring caveat that a `None`-seed `UMAPResult` is explicitly non-reproducible, given the PR's own emphasis on UMAP's seed being load-bearing.
- Minor: the producer's `perform_umap_analysis(random_state: int = 42)` type annotation doesn't reflect that `None` is an accepted (if non-deterministic) input; `Optional[int]` would be more accurate. Cosmetic, pre-existing pattern, not introduced by this PR's core change.

---

PR #182, opened 2026-07-09
feat(#179): hierarchical clustering entry point returning a labeled ClusterResult

egao28:
## Summary

Adds a public hierarchical-clustering entry point that returns a labeled, JSON-serializable `ClusterResult` — the follow-up to #129 that `bloom-mcp` needs. Previously `perform_hierarchical_clustering` returned only a dendrogram (no cluster labels), so a `ClusterResult` couldn't be built from it.

Implements all three requirements from #179:

1. **`hierarchical_cluster_labels(data, *, n_clusters=None, method="ward", metric="euclidean", standardize=True, optimization_method="silhouette", max_clusters=10)`** — a public producer in `clustering.py` that **composes existing internals** (`perform_hierarchical_clustering` → `calculate_optimal_clusters_hierarchical` when `n_clusters` is omitted → `cut_dendrogram`) into one labeled dict: cluster labels/sizes, the three quality metrics, hierarchical provenance (`linkage_method`, `distance_metric`, `cophenetic_correlation`, `cut_height`), and `data_indices` (mapping labels back to source rows after NaN-dropping). It does **not** change `perform_hierarchical_clustering`'s return shape.
2. **`HierarchicalResult`** (frozen `ClusterResult` subclass) + **`ClusterResult.from_hierarchical_dict(d)`** adapter (no `random_state` arg — hierarchical is deterministic — stamps `None`) + the **`ALGORITHM_HIERARCHICAL`** discriminator constant.
3. **`ClusterResult.random_state` → `Optional[int]` (default `None`)**, matching `PCAResult.random_state`.

Acceptance criteria met: a single public call returns a populated `HierarchicalResult`; `ClusterResult(random_state=None)` is valid; identical input → identical labels.

## Design notes

- **Producer → dict → adapter split preserved.** A new producer + `from_hierarchical_dict` mirrors the existing `from_kmeans_dict`/`from_gmm_dict` convention exactly (the adapter only maps, it does not compute), and keeps `perform_hierarchical_clustering` dendrogram-only so its existing consumers are unaffected.
- **`random_state` widening** is source-compatible for producers (KMeans/GMM still stamp the `int` seed) but a reader that assumed an always-`int` value must now handle `None`. To prevent the widening from silently letting a seeded subclass construct without a real seed, `KMeansResult`/`GMMResult` reject `random_state=None` via `__post_init__` (`TypeError`) — only `HierarchicalResult` may omit it. The downstream schema is updated in bloom-mcp.
- **`ALGORITHM_HIERARCHICAL` is exported from `result_types` only**, not the package root — the root public-API docstring audit rejects bare `str` constants in `__all__` (same treatment as `ALGORITHM_KMEANS`/`ALGORITHM_GMM`).
- **No version bump here.** Per #176, `0.1.0a5` is cut in a separate `chore` release PR via `uv version`; this PR carries only a `docs/CHANGELOG.md` `[Unreleased]` entry.
- **Error contract:** every invalid *argument* (`method`, `metric`, `optimization_method`, non-integer/out-of-range `n_clusters`) raises `ValueError` — validated up front, or propagated from the composed functions, so a caller catches one exception type. A genuine *data* failure (e.g. non-numeric columns reaching `linkage` with `standardize=False`) is intentionally left as `RuntimeError` — that's a bad input, not a bad argument.
- **Determinism caveat, stated explicitly:** no RNG in the composed call path, so identical input yields identical labels **within one process**; scipy's `linkage`/`fcluster` tie-breaking is BLAS/platform-sensitive, so this is not a promise of cross-platform reproducibility — no golden artifact is committed for it.

## Review history

This PR went through two rounds of review (both fixed, pushed, and re-verified against a full test run before each push):

- **Round 1** caught a mixed `ValueError`/`RuntimeError` contract (`method`/`metric` reaching scipy unvalidated) and a dropped `data_indices` — fixed in `fb0fdbd`/`5c03a10`.
- **Round 2** (a 5-lens subagent team plus manual review) caught one confirmed **blocking** bug — `method="centroid"`/`"median"` + a non-euclidean metric still leaked `RuntimeError` (scipy requires euclidean for all three of `ward`/`centroid`/`median`, not only `ward`) — plus four Important-tier gaps: a reproducibility guardrail regression on `KMeansResult`/`GMMResult` (now restored), an unqualified determinism claim in the docs, an under-tested `feature_names` footgun (documented, pre-existing/epic-wide), and a real ~11% perf redundancy in the auto-`k` path (fixed). All fixed in `a988210`.

## Test plan

- `tests/test_cluster_result.py`: 37 tests total (13 hierarchical-specific: round-trip, `Optional` `random_state` + the KMeans/GMM guardrail, adapter mapping + non-mutation, `to_json` finite guard incl. a real-degenerate-input NaN case, exports).
- `tests/test_clustering.py`: 88 tests total (31 in the new `TestHierarchicalClusterLabels` class — auto/explicit `k`, optimizer delegation, determinism, the `optimization_method` value-set + footgun, argument-validation edge cases (`bool`/`numpy.float64`/`<=0` `n_clusters`, unknown `method`/`metric`), single-cluster finite metrics, `data_indices` mapping, end-to-end typed view — plus 2 centroid/median regression tests added to the existing `TestHierarchicalEdgeCases`).
- **Full suite: 2461 passed, 31 skipped, 3 deselected** (integration) — green.
- Black clean; Ruff clean on source. Public-API docstring audit passes.

This change was developed spec-first via the OpenSpec change `add-hierarchical-cluster-result` (proposal + design + tasks + delta spec included), which was reviewed by a 5-agent adversarial review before implementation and updated in lockstep with both review rounds above.

Closes #179.


eberrigan (changes requested):
## Review Summary

Clean, convention-respecting addition — the producer → dict → adapter split mirrors `from_kmeans_dict`/`from_gmm_dict` exactly, the `random_state` → `Optional[int]` widening is safe (no `src/` reader assumes `int`, dataclass ordering valid, no existing test breaks), all 8 OpenSpec scenarios are faithfully implemented and 1:1 mapped to tests, and docs are consistent with no drift. Cross-platform-safe and no OOM regression (`linkage`/`pdist` run once and are reused; tests assert only structural bounds + in-process equality, no committed goldens).

Two substantive items should be addressed before merge: the "every argument error → `ValueError`" contract is not actually honored, and the labeled dict silently drops sample-to-row provenance.

## Blocking Issues

**1. The "every argument error → `ValueError`" contract is not honored for `method` / `metric` / non-int `n_clusters`.**
This is the PR's stated purpose (design.md: *"Every invalid argument surfaces as ValueError, so a consumer (bloom-mcp) catches one exception type"*; commit `fb0fdbd`). But only the ward+euclidean *combination* and `optimization_method`/`n_clusters`-range are validated up front. A bogus `method`/`metric` reaches `linkage`/`pdist` **inside** `perform_hierarchical_clustering`'s try/except (`clustering.py:574-601`), which re-wraps scipy's `ValueError` as **`RuntimeError`**:
- `hierarchical_cluster_labels(df, method="bogus")` -> `RuntimeError`
- `hierarchical_cluster_labels(df, method="average", metric="bogus")` -> `RuntimeError`
- `hierarchical_cluster_labels(df, n_clusters=1.5)` -> `RuntimeError` (`1 <= 1.5 <= n-1` passes the range check, then a float index into the linkage matrix throws)

A bloom-mcp wrapper catching only `ValueError` will leak these. **Fix:** validate `method`/`metric` against the accepted sets and require integer `n_clusters` up front — or narrow design.md/the docstring to state exactly which argument errors are normalized.

**2. The labeled dict drops `data_indices`, unlike every sibling producer — silent sample-to-row traceability loss on NaN input.**
`perform_kmeans_clustering`, `perform_gmm_clustering`, and `cut_dendrogram` all return `data_indices` so labels map back to original rows after NaN-row dropping. `hierarchical_cluster_labels` omits it even though `cut["data_indices"]` is in scope (`clustering.py:721-733`). With partial-NaN input, `cluster_labels` aligns to the *surviving* rows with no way to recover the original mapping, and the docstring doesn't mention the drop. For root phenotyping (labels -> plant/genotype IDs) that mapping is scientifically load-bearing. **Fix:** add `"data_indices": cut["data_indices"]` to the returned dict (the typed `ClusterResult` family has never carried it, so `HierarchicalResult` needn't — this is purely a producer-dict fix), and add a partial-NaN test.

## Important Issues

- **Davies-Bouldin `0.0` for a single cluster is actively misleading** (`clustering.py:680-685`). For DB, lower is better, so `0.0` is the *best possible* score — a degenerate `n_clusters=1` run reports a "perfect" DB score for an undefined metric; a consumer ranking runs by DB would pick the degenerate one. Document the asymmetry, or return `NaN` (the design chose `0.0` for `to_json` serializability).
- **Cophenetic `NaN` surfaces only at `to_json`, not at production.** Degenerate inputs (a 2-row input, or `standardize=False` on identical points) yield `cophenetic_correlation=NaN`, accepted silently by the producer and adapter; it only detonates at `.to_json()`. A consumer using the dict/dataclass directly gets a `NaN` "quality metric" with no signal. Surface it earlier or document it. (`standardize_data` itself is safe — it drops zero-variance columns and raises if all are constant.)
- **`optimization_method` is silently ignored when `n_clusters` is set** (`clustering.py:696` — guarded only `if n_clusters is None`). Validate unconditionally or document "ignored when n_clusters is set."
- **The accepted-methods set is duplicated** — inlined in the producer (`clustering.py:695`) and independently in `calculate_optimal_clusters_hierarchical` (`clustering.py:750-760`). If one drifts, a dropped method passes the producer check then hits the optimizer's `else` -> `RuntimeError`, reintroducing item #1. Hoist a shared module-level constant.
- **Test gaps / brittleness:** partial-NaN input is untested, and 4 of 6 error-path tests `match` on messages owned by the *composed internals* (reword an internal message and the producer tests break). Assert exception *type* for the propagated cases; the internal messages are already pinned by their own direct tests.
- **Auto-`k` recomputes silhouette redundantly** (`clustering.py:706-719`): the optimizer runs 9 O(n²) silhouette passes (k=2..10), then the producer re-cuts for the winning `k` -> 1 more, fully redundant O(n²) pass. Bounded by the O(n²) linkage ceiling and acceptable given the "compose unchanged internals" goal, but worth a follow-up note.

## Suggestions

- `feature_names` is captured before `standardize_data` drops zero-variance columns, so it can be longer than the feature count actually clustered — pre-existing and inherited by all producers, but now exposed in a typed result. Worth a separate issue.
- Add explicit tests pinning the producer's `cluster_labels` type (numpy array), the empty-DataFrame path, and single-row auto-`k`.
- A one-line docstring note on the practical sample-size ceiling (O(n²) memory) for this now-public entry point.

---

eberrigan (changes requested):
## Review Summary

Solid, spec-first addition (`hierarchical_cluster_labels()` + `HierarchicalResult` + `from_hierarchical_dict`) that follows the existing producer→dict→adapter convention cleanly, with strong test coverage and good docstrings. However, one real bug survived the PR's own §5 review-follow-up pass: two of the seven accepted linkage methods still leak a `RuntimeError` where the PR explicitly promises `ValueError` for every argument error — confirmed independently by three separate reviewers (including live execution against the PR branch), so this needs a fix before merge. A handful of Important-tier gaps (a reproducibility guardrail regression, an unqualified determinism claim, an under-tested `feature_names` footgun, and a real minor perf redundancy) are worth addressing too but are not merge-blocking on their own.

## Blocking Issues

**1. `method="centroid"` or `method="median"` + a non-euclidean metric leaks `RuntimeError`, not `ValueError`** — violates this PR's own stated contract.

`perform_hierarchical_clustering` (`src/sleap_roots_analyze/clustering.py`) only pre-checks the `ward`+non-euclidean combination before its `try` block:
```python
if method == "ward" and metric != "euclidean":
    raise ValueError("Ward linkage requires euclidean metric")
```
scipy's `linkage()` enforces the same euclidean-only constraint for `centroid` and `median`, but neither `perform_hierarchical_clustering` nor the new up-front validation in `hierarchical_cluster_labels` (which only checks `method`/`metric` set-membership individually, `clustering.py:909-927`) catches this combination. The call reaches `linkage()` inside the existing `try/except`, which rewraps the genuine `ValueError` scipy raises into `RuntimeError` (`clustering.py:639-640`).

Verified live in an isolated worktree on the PR branch:
```
hierarchical_cluster_labels(df, method="centroid", metric="cosine")
-> RuntimeError: Hierarchical clustering failed: `method=centroid` requires the distance metric to be Euclidean
hierarchical_cluster_labels(df, method="median", metric="cosine")
-> RuntimeError: Hierarchical clustering failed: `method=median` requires the distance metric to be Euclidean
```
This directly contradicts `design.md`, `proposal.md`, the `spec.md` "Invalid input raises ValueError" scenario, the docstring's `Raises:` section, and task 5.1's explicit goal ("Honor the ... `ValueError` contract fully"). A consumer (bloom-mcp) catching only `ValueError` for argument errors, as promised, will get an uncaught `RuntimeError` for these two methods. No test exercises `method="centroid"`/`"median"` at all — `test_ward_non_euclidean_raises_value_error` only covers `ward`.

**Fix:** extend the pre-`try` constraint check to `method in {"ward", "centroid", "median"} and metric != "euclidean"` (either in `perform_hierarchical_clustering` or as an additional up-front check in `hierarchical_cluster_labels`), update the docstring's `Raises:` wording accordingly, and add regression tests for `centroid`/`median` mirroring the existing `ward` test.

## Important Issues

**2. `ClusterResult.random_state: Optional[int] = None` silently removes a reproducibility guardrail for `KMeansResult`/`GMMResult`.** Before this PR, omitting `random_state` on direct construction raised `TypeError`. Now `KMeansResult(...)`/`GMMResult(...)` silently succeed with `random_state=None`, even though these algorithms always have a real seed via their producers. Since both classes are root-exported public API, consider a `__post_init__` on `KMeansResult`/`GMMResult` that rejects `random_state=None` — only `HierarchicalResult` should accept it. Related: no test anywhere asserts `GMMResult.random_state` is actually an `int` (the `KMeansResult` equivalent is tested), so this exact regression path is currently unverified for GMM.

**3. The "hierarchical clustering is deterministic" claim is unqualified in shipped docstrings/docs.** `design.md`/`tasks.md` note that scipy's `linkage`/`fcluster` tie-breaking is "BLAS/platform-sensitive" and that is why no golden artifact is committed — but that caveat never made it into the `HierarchicalResult`/`hierarchical_cluster_labels` docstrings or `docs/result-types.md`, which flatly say "deterministic." A consumer reading only the public docs would reasonably assume cross-platform reproducibility. Worth a one-line caveat in the public-facing docstrings, especially since this PR makes the auto-k (tie-break-sensitive) path the default behavior of a single convenient call — previously it required manually chaining three functions to reach.

**4. `feature_names` can silently list columns that were never used for clustering.** `standardize_data` (pre-existing, shared by kmeans/gmm/hierarchical) drops non-numeric and zero-variance columns, but `feature_names` is captured before that call. Verified live: a 5-column DataFrame (one non-numeric, one constant) reports all 5 names in `feature_names` while `data_processed` only has 3 columns. This is pre-existing behavior, not a regression, but it is newly exposed via the new public entry point and the new `HierarchicalResult.feature_names` field, which `design.md` calls scientifically load-bearing provenance — and its own Risks section never mentions this. Worth a follow-up issue.

**5. Non-numeric column + `standardize=False` leaks `RuntimeError`, not `ValueError`.** `hierarchical_cluster_labels(df_with_string_col, standardize=False)` raises `RuntimeError: ... could not convert string to float`. Arguably "bad data, not a bad argument," but inconsistent with the PR's "every argument error is ValueError" framing, and untested.

**6. Redundant `cut_dendrogram` re-execution on the default auto-k path.** When `n_clusters=None` (the default), `calculate_optimal_clusters_hierarchical`'s internal scan already computes `cut_dendrogram(dendrogram, n_clusters=optimal_k)` for the winning `k`, but `hierarchical_cluster_labels` then unconditionally re-runs `cut_dendrogram` for that same `k` to get the labels — an avoidable ~1/(max_clusters−1) (~11% with default `max_clusters=10`) re-execution of an O(n^2) silhouette computation, by default. Confirmed the underlying `linkage`/`pdist` step itself is not duplicated (computed once, correctly reused) — only the final cut+metrics step is. Fixable by threading the winning `cut_result` out of the optimizer's loop instead of discarding it.

**7. `davies_bouldin_score == 0.0` for `n_clusters=1` is a footgun beyond documentation.** Lower is better for Davies-Bouldin, so `0.0` reads as the best possible score for an undefined metric. The auto-k path never selects `k=1` so it is safe there, but nothing stops a consumer from ranking manually-constructed `HierarchicalResult`s (including `n_clusters=1` ones) by this score and picking the degenerate one as "best." Documented in the docstring, which helps, but consider treating it like `cophenetic_correlation` (propagate `NaN`, rejected only at `to_json`) for consistency, or exposing an `is_degenerate` signal.

**8. PR description test counts appear stale.** Claimed "16 in `test_cluster_result.py`, 14 in `test_clustering.py`"; actual collected counts on the PR head are 14 and 24 respectively (the "14" matches only the first implementation commit, before the two review-follow-up commits added more). Minor, but worth fixing before merge since the description is otherwise very precise.

**9. Missing edge-case tests:** boolean `n_clusters` (an `isinstance(n_clusters, bool)` guard exists at the new validation code but no test exercises it) and `n_clusters <= 0` (only the upper-bound `n_clusters == n_samples` case is tested, not the lower bound).

## Suggestions

- Short-circuit the `n_clusters` out-of-range check before paying for the O(n^2) `linkage`/`pdist` step, where feasible.
- Add the O(n^2)-memory docstring note to `perform_hierarchical_clustering` itself, not only the new wrapper — it is the function actually doing that work.
- Add an end-to-end (non-synthetic) test that drives a real `NaN` `cophenetic_correlation` through the full pipeline (currently only tested via synthetic `dataclasses.replace`).
- Add a `numpy.float64` case alongside the existing plain-`float` `n_clusters`-type-rejection test.
- Validate `max_clusters >= 2` up front regardless of whether `n_clusters` is explicit, for a clearer error message.
- Flag the `feature_names`/standardize footgun and `data_indices` absence from the typed view more prominently in `docs/result-types.md`.

---
*Review by Claude Code subagent team (Code Quality, Testing, Statistical Rigor, Performance/Memory, Behavioural Correctness). Findings 1-9 above were each independently corroborated by at least one subagent doing live code execution against the PR branch, not static reading alone.*

eberrigan (approved):
## Review Summary (round 2)

Re-reviewed after commit `a988210` ("fix(#179): fix centroid/median ValueError leak, restore random_state guardrail"). The blocking issue and the key Important-tier gaps from round 1 are resolved — verified live on the updated branch, not just by reading the diff. Approving.

## Verified fixes

- **Blocking #1 (`centroid`/`median` + non-euclidean metric leaking `RuntimeError`) — fixed.** Re-ran live: both now correctly raise `ValueError` ("Centroid linkage requires euclidean metric" / "Median linkage requires euclidean metric"). The fix is applied at the single source of truth in `perform_hierarchical_clustering`, so it also covers `detect_outliers_hierarchical`/`create_dendrogram`, not just the new producer. Regression tests added at both levels.
- **Important #2 (`random_state` guardrail regression) — fixed.** Re-ran live: constructing `KMeansResult(..., random_state=None)` now correctly raises `TypeError` via the new `__post_init__`; only `HierarchicalResult` may omit it. The missing `GMMResult.random_state` int-type assertion was also added.
- **Important #3 (unqualified determinism claim) — fixed.** Docstrings and `docs/result-types.md` now distinguish same-process determinism from cross-platform reproducibility, consistent with the existing no-golden-artifact design decision.
- **Important #6 (redundant `cut_dendrogram` re-execution on auto-k) — fixed.** `calculate_optimal_clusters_hierarchical` now returns the winning candidate's `cut_result` from its scan, and the producer reuses it instead of re-cutting — the extra ~11% O(n^2) pass is gone.
- **Important #9 (missing edge-case tests) — fixed.** `bool`, `numpy.float64`, and `n_clusters <= 0` cases are now covered, plus a genuine end-to-end NaN-`cophenetic_correlation` test via all-identical input (not just a synthetic `dataclasses.replace`).

Ran the full `tests/test_cluster_result.py` + `tests/test_clustering.py` suite on the updated branch: **125 passed** (up from 111 in round 1).

## Remaining non-blocking notes

- **Important #4 / #5** (the pre-existing `feature_names` footgun when `standardize_data` drops columns, and `standardize=False` + a non-numeric column raising `RuntimeError` rather than `ValueError`) were triaged rather than code-changed — both are now documented in `docs/result-types.md`, and #5 has a regression test locking in the intentional behavior. That's a reasonable call: #5 is genuinely a data failure rather than an argument error, and #4 is a pre-existing, epic-wide pattern shared by `KMeansResult`/`GMMResult` that shouldn't be fixed piecemeal in this PR. Worth a follow-up issue if not already tracked.
- **Important #7** (`davies_bouldin_score == 0.0` for a single cluster reading as the "best" score) is still only documented, not code-changed — acceptable as-is given the auto-k path never selects `k=1`, but still a footgun for a consumer manually constructing/ranking `HierarchicalResult`s.
- **Important #8** (stale test counts in the PR description) doesn't appear to have been updated — minor, worth a quick edit to the PR body before merge but not blocking.

No new issues found. Nice, fast turnaround on the fix.

---
*Review by Claude Code (round 2 follow-up to the subagent-team review). Verified live against the updated PR branch.*

PR #185, opened 2026-07-13
fix(#183): re-derive clustering feature_names from filtered columns

egao28:
## Summary
- `perform_kmeans_clustering`, `perform_gmm_clustering`, and `perform_hierarchical_clustering` snapshotted `feature_names` **before** `standardize_data` dropped non-numeric/zero-variance columns, silently mislabeling `cluster_centers`/`means`/`data_processed` for such inputs — not a length mismatch, a positional mislabeling. Ports the pattern already correct in `perform_pca_analysis` (`pca.py:793-812`) to both `standardize` branches; `standardize=False` now applies the same numeric + non-zero-variance filter `standardize=True` always used internally, instead of no filtering at all.
- `KMeansResult`/`GMMResult` and `detect_outliers_kmeans`/`_gmm`/`_hierarchical` (which dict-spread the producer output and had zero `feature_names` coverage) inherit the corrected values automatically — no adapter code changes needed, but added regression tests since nothing previously covered them.
- No breaking changes: dict keys, shapes for already-clean input, and typed adapter signatures are unchanged. Full OpenSpec proposal at `openspec/changes/fix-clustering-feature-names-mismatch/`.

Closes #183.

**Note on #179/PR #182:** the new `HierarchicalResult`/`hierarchical_cluster_labels` (PR #182) also inherit this bug once merged, but PR #182 is still open — this PR only touches `clustering.py`'s three producers, which are independent of it. Whichever PR merges second should add a small follow-up test confirming `HierarchicalResult.feature_names` reflects the corrected values (no code change expected).

## Test plan
- [x] New parametrized tests across all three producers × both `standardize` values: `feature_names` length matches the array, constant/non-numeric columns excluded in original order, named-value regression against `data_processed` (catches mislabeling a length check alone would miss), fully-filtered-input error path
- [x] `KMeansResult`/`GMMResult` adapter regression coverage
- [x] `detect_outliers_kmeans`/`_gmm`/`_hierarchical` regression coverage (previously zero `feature_names` coverage)
- [x] Pipeline-path guardrail in `DetectOutliersStep`
- [x] Full suite: 2486 passed, 31 skipped, 0 failed
- [x] `ruff check`, `black --check`, and the frozen `mypy` baseline (0 new errors) all clean

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -32,6 +32,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   already removed earlier by the zero-inflation filter, so only constant-nonzero traits are
   affected.)
 
+### Fixed
+- `perform_kmeans_clustering`, `perform_gmm_clustering`, and
+  `perform_hierarchical_clustering` (`clustering.py`) now re-derive `feature_names` from
+  the columns actually used for fitting, on both `standardize=True` and
+  `standardize=False` (#183). Previously `feature_names` was snapshotted **before**
+  constant/non-numeric columns were filtered out, silently mislabeling
+  `cluster_centers`/`means`/`data_processed` for such inputs — not a length mismatch a
+  caller would notice, a positional mislabeling. `standardize=False` additionally now
+  applies the same numeric + non-zero-variance filter `standardize=True` always used
+  internally; previously it applied no filtering at all, so a non-numeric column reached
+  the estimator directly and failed with a raw sklearn "could not convert string to float"
+  error instead of the clear "No numeric columns with non-zero variance found" message.
+  This is separate from the `#177` cleanup change above: `#177` keeps constant traits from
+  ever reaching these functions through the standard cleanup step; `#183` fixes the
+  functions' own label bookkeeping for any caller that doesn't go through it — e.g. direct
+  callers, or `standardize=False`. `KMeansResult`/`GMMResult` and
+  `detect_outliers_kmeans`/`_gmm`/`_hierarchical` (`outlier_detection.py`) inherit the
+  corrected values automatically, with no adapter code changes needed.
+
 ## [0.1.0a4] - 2026-07-02 (Pre-release)
 
 ### Added
--- a/openspec/changes/fix-clustering-feature-names-mismatch/proposal.md
+++ b/openspec/changes/fix-clustering-feature-names-mismatch/proposal.md
@@ -1,0 +1,145 @@
+## Why
+
+`perform_kmeans_clustering`, `perform_gmm_clustering`, and
+`perform_hierarchical_clustering` (all in `clustering.py`) each snapshot
+`feature_names = df_clean.columns.tolist()` **before** calling
+`standardize_data`, then return that snapshot unchanged even when
+`standardize_data` goes on to drop non-numeric and exact-zero-variance
+columns (`pca.py:624-654`). The result is a silent **positional**
+misalignment between `feature_names` and every feature-indexed array in the
+same return dict (`cluster_centers`, `means`, `data_processed`) — not a
+length mismatch a caller would notice, a mislabeling. All three producers
+discard `standardize_data`'s filtered `df_clean` via `_` (e.g.
+`X_processed, scaler, _ = standardize_data(df_clean)`), so nothing in the
+`standardize=True` branch ever re-derives names from what was actually
+clustered. The `standardize=False` branch is worse: it does no filtering at
+all (`X_processed = df_clean.values`), so a non-numeric column reaches
+`KMeans.fit()` / `GaussianMixture.fit()` / `linkage()` directly and either
+mislabels `feature_names` (constant column) or fails with sklearn's own
+`could not convert string to float` error (non-numeric column) instead of
+the clearer "No numeric columns with non-zero variance found" message the
+`standardize=True` path already gives for the equivalent all-filtered case.
+(Verified empirically: both cases are already wrapped in a method-specific
+`RuntimeError` by these producers' existing broad `except Exception`
+handlers — this is not an uncontrolled crash today, just a confusing
+message and, independently, a `feature_names` mismatch whenever at least
+one valid column survives.)
+
+This is the same bug class already fixed for PCA in #74/#76:
+`perform_pca_analysis` (`pca.py:793-812`) re-derives `feature_names` from the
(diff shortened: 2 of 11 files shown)


egao28 (commented):
> **Verdict: COMMENT** (posted as comment — cannot approve/request-changes on your own PR)

## Review Summary

This is a precise, well-scoped fix for #183: `perform_kmeans_clustering`/`perform_gmm_clustering`/`perform_hierarchical_clustering` snapshotted `feature_names` before `standardize_data` dropped non-numeric/zero-variance columns, silently mislabeling `cluster_centers`/`means`/`data_processed`. Five independent review passes (code quality, testing/TDD, statistical rigor, performance/cross-platform, behavioral correctness) verified the fix by reading `clustering.py` line-by-line and diffing it against the `perform_pca_analysis` pattern it ports (`pca.py:793-812`), plus live execution of the issue's exact repro and several pathological inputs. **No blocking issues.** All CI checks are now green (Tests on ubuntu/mac/windows, Lint, mypy, numerical-stability, reproducibility, serialization round-trip).

Positional alignment is correct in both `standardize` branches for all three producers; `original_indices` (row-level) is unaffected since `standardize_data` only filters columns; the new `ValueError` on `standardize=False`'s all-filtered case is correctly caught and re-wrapped as `RuntimeError`, matching the `standardize=True` path's existing contract; downstream inheritance into `KMeansResult`/`GMMResult` and `detect_outliers_kmeans`/`_gmm`/`_hierarchical` was confirmed correct by reading the actual dict-spread/pass-through code, not just trusting the PR description. `outlier_detection.py` and `result_types.py` have zero production diffs, consistent with the "inherits automatically" claim.

## Important Issues

1. **Named-value regression test doesn't exercise `array_key`.** `test_named_value_matches_processed_data` (`tests/test_clustering.py:729`) is parametrized over `array_key` (`cluster_centers`/`means`/`data_processed`) but the body always indexes `result["data_processed"]`, never `result[array_key]`. So `cluster_centers` (KMeans) and `means` (GMM) are never independently checked at the named-value level — only structurally (length) and via the shared `data_processed` proxy. Functional risk is low (sklearn preserves the fit-input column order, so if `data_processed` is aligned, `cluster_centers_`/`means_` are too), but `tasks.md` Task 3.1 is checked off as if the `cluster_centers[k]`/`means[k]` check from the issue's test plan exists — it doesn't quite. Worth either adding the `array_key`-specific assertion or correcting the task-completion note.
2. **Commit order doesn't match the stated TDD plan.** `tasks.md` describes a red-then-green sequence (tests fail → fix commit turns them green), but `a2246f1` (fix, 0 test files touched) landed 14s before `6f35385` (tests, 0 source files touched) in the branch history. The tests are genuinely real regressions — reverting `clustering.py` and rerunning them confirms 39 failures pre-fix — so this isn't a tautological-test problem, just a discrepancy between the documented process and the actual commit sequence.

## Suggestions

- The numeric-dtype + non-zero-variance filter block (`select_dtypes` → `var(ddof=0)` → mask) is now duplicated 3x in `clustering.py`, on top of 2 existing copies in `pca.py` — a shared helper would collapse all 5 and prevent future drift.
- Minor pre-existing asymmetry (not introduced here): `pca.py`'s equivalent all-filtered condition raises a bare uncaught `ValueError`, while `clustering.py` wraps the same condition into `RuntimeError` via its `try`/`except`. Not a defect, just inconsistent between the two modules.
- No fixture pairs a non-numeric column *without* a constant column (the mixed fixture always includes both). Live execution confirms this path works correctly for all 3 producers × both `standardize` values, but nothing in the suite pins it down directly.
- Consider one explicit "byte-identical to pre-fix output on already-clean input" pinned test rather than relying on the filter being provably a no-op by inspection.

---

eberrigan (commented):
> **Verdict: COMMENT**

## Review Summary

This is a well-scoped, correctly-implemented bug fix. All three clustering producers (`perform_kmeans_clustering`, `perform_gmm_clustering`, `perform_hierarchical_clustering`) previously snapshotted `feature_names` before column filtering, causing a real positional mislabeling bug on `standardize=True` for inputs with constant/non-numeric columns. The fix correctly ports the existing `perform_pca_analysis` pattern, adds a shared `filter_numeric_nonzero_variance()` helper, and includes strong regression tests (named-value mapping, not just length checks). Black/ruff/mypy/pytest verified clean in a local worktree; CI is green on all 3 OSes.

## Blocking Issues

None.

## Important Issues

1. **TDD claim in the PR description doesn't match the commit history.** The PR body claims tests were written first (red phase) before the fix (green phase). Actual commit timestamps show the fix commit (`a2246f1`) landed ~14 seconds *before* the test commit (`6f35385`) — there's no evidence the tests ever failed against pre-fix code. This doesn't affect code correctness, but the description should be corrected so it doesn't misrepresent the process.

2. **`standardize=False` behavior change is described as "the same bug" as the `standardize=True` mislabeling, but it isn't quite.** Before this PR, `standardize=False` was internally *consistent* — `feature_names` and `X_processed` were both derived from the same unfiltered `df_clean`, so there was no positional mislabeling in that branch (unlike `standardize=True`, where the real bug lived). What this PR does for `standardize=False` is add brand-new filtering — a behavior/validation change, not a mislabeling fix. Practical consequence: any existing caller using `standardize=False` on data with a constant column will get a **different array shape** for `cluster_centers`/`means`/`data_processed` (one fewer column) after upgrading, not just a corrected label. The current CHANGELOG wording is general enough to imply this is the same silent-mislabeling bug for both branches. Recommend tightening the CHANGELOG/proposal to explicitly call out the `standardize=False` shape change as a new, deliberate validation addition rather than folding it into the mislabeling narrative.

## Suggestions

- `filter_numeric_nonzero_variance` isn't added to any `__all__`/export list — consistent with existing convention (`standardize_data` isn't either), so not a new inconsistency, just noted.
- `calculate_optimal_k_kmeans` still discards `standardize_data`'s filtered `df_clean` via `_` — harmless today (it never returns `feature_names`) but worth a one-line comment if a future change extends it.
- Consider a debug-level log when columns are silently dropped under `standardize=False`, since there's currently no signal besides inspecting `feature_names` length.

---

eberrigan (approved):
No blocking issues from the subagent review team (code quality, testing/TDD, statistical rigor, performance/cross-platform, behavioral correctness) — see prior comment for the two non-blocking documentation items worth a quick follow-up. CI is green across Ubuntu/Windows/macOS.

Issue #109, opened 2026-04-03
Add BLUP extraction to heritability analysis for shrinkage-adjusted genotype means


eberrigan:
## Problem

The QC pipeline fits a linear mixed model (`statsmodels MixedLM` with REML) to estimate broad-sense heritability (H²), but discards the fitted model's random effects. These random effects are BLUPs (Best Linear Unbiased Predictions) — shrinkage-adjusted genotype estimates that are more reliable than raw genotype means, especially for unbalanced designs.

Currently, downstream analysis (PCA, UMAP, clustering) operates on raw genotype means, which do not account for unequal replication. Genotypes with few replicates get noisy estimates that can dominate multivariate projections.

## Background

BLUPs are the standard method for ranking genotypes in plant breeding and phenomics. They are computed as the conditional modes of the random effects from the same mixed model we already fit for heritability. In statsmodels, this is `result.random_effects` — mathematically equivalent to R's `lme4::ranef()`.

For each genotype, the adjusted mean (BLUE + BLUP) is:

```
adjusted_mean[g] = intercept + blup[g]
```

where `intercept = result.fe_params["Intercept"]` and `blup[g] = result.random_effects[g][0]`.

Genotypes with fewer replicates are "shrunk" toward the grand mean, reducing noise. With balanced designs and many reps, BLUPs converge to raw means — so this is strictly more general.

### 1. Extract BLUPs during heritability calculation

In `statistics.py:calculate_heritability_estimates()`, after fitting the mixed model (line ~293), extract:

- `result.random_effects` — BLUP for each genotype (dict)
- `result.fe_params["Intercept"]` — fixed-effect intercept (grand mean)
- `result.random_effects_cov` — posterior covariance (optional, for reliability)

Return these in the per-trait results dict alongside existing `heritability`, `var_genetic`, `var_residual` fields.

### 2. Add a function to produce a genotype × trait BLUP matrix

New function `extract_blup_table()` that takes heritability results and returns a `pd.DataFrame` with genotypes as rows and traits as columns, containing adjusted means (intercept + BLUP). This is the standard input for downstream multivariate analysis.

### 3. Integrate BLUP table into pipeline output

The `StatisticalAnalysis` step should optionally write a `blup_adjusted_means.csv` alongside existing outputs. Controlled by a config flag (default: enabled, since it's free once the model is fit).

### 4. No new dependencies

`statsmodels` already computes BLUPs internally — we just need to extract `result.random_effects`. No new packages required.

[...]

PR #189, opened 2026-07-15
feat(#109): BLUP-adjusted genotype means (Tier 1 cross-platform prediction)

eberrigan:
## Summary

Tier 1 of the cross-platform genotype-prediction program (#109): extracts BLUP (Best Linear Unbiased Prediction) adjusted genotype means from the existing heritability mixed-model fit. BLUP-adjusted means are the predictor substrate later tiers use to test whether one phenotyping platform's genotype effects predict another's.

- `calculate_heritability_estimates` additively returns `blup`/`intercept` per trait when its mixed model succeeds (both existing return shapes unchanged).
- New `extract_blup_table(heritability_results) -> pd.DataFrame`: genotype x trait BLUP-adjusted-means table, with genuine `NaN` (not zero-filled, not dropped) for any trait/genotype that didn't succeed.
- New `BLUPResult` dataclass + `BLUPResult.from_blup_table(df, *, intercepts=None)` adapter — frozen, JSON-serializable, mirrors `HeritabilityResult`'s per-trait/failed_traits split one level up.
- New `StatisticsConfig.generate_blup_table` (default `True`) — pipeline writes `08_blup_adjusted_means.csv` alongside `08_heritability_results.csv` when both it and `calculate_heritability` are enabled; setting it `True` while `calculate_heritability=False` is a silent no-op (see `design.md` Decision 5).
- `extract_blup_table` + `BLUPResult` exported from the package root.

OpenSpec proposal: `openspec/changes/add-blup-extraction/` (validated `--strict`).

### Fixes from pre-PR self-review

A 5-agent adversarial review (code quality, testing, statistical rigor, performance/cross-platform, behavioural correctness) independently surfaced two real bugs, both fixed on this branch before opening the PR:

1. **Genotype identity was silently dropped from `08_blup_adjusted_means.csv`.** `extract_blup_table()` returns a genotype-indexed DataFrame, but the pipeline's shared `save_dataframe()` writes with `index=False` — the CSV had no genotype column at all. Fixed by resetting the index into a named `Genotype` column before saving; strengthened the regression test to assert the column and its values are present, not just row count.
2. **`BLUPResult.adjusted_means` broke its documented `(n_genotypes, n_traits)` shape** when every trait failed via a cell-level gap (not a zero-row DataFrame) — `genotype_names` stayed populated but `adjusted_means` collapsed to `[]` instead of one empty row per genotype. Fixed by removing an unnecessary/wrong special case; added a regression test.

Also added: the missing `calculate_heritability=False, generate_blup_table=False` config combination test, and a trivial unnecessary-quote cleanup on a type hint.

## Test plan

- [x] `black --check` / `ruff check` clean
- [x] `openspec validate add-blup-extraction --strict` → valid
- [x] Full suite: `uv run pytest --cov=src/sleap_roots_analyze --cov-report=term-missing tests/` → 2617 passed, 31 skipped, 0 failed
- [x] `result_types.py` (new `BLUPResult`) at 100% coverage; `pipeline/steps/statistical_analysis.py` 94%, `pipeline/config/components.py` 99%
- [x] 5-subagent pre-PR self-review completed; both BLOCKING findings fixed and covered by new regression tests

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -591,6 +591,41 @@ Build a comparison table of variance components and heritability across traits.
 
 ---
 
+#### `extract_blup_table`
+
+```python
+extract_blup_table(
+    heritability_results: Dict
+) -> pd.DataFrame
+```
+
+Build a genotype x trait BLUP-adjusted-means table from a
+`calculate_heritability_estimates` result (issue #109).
+
+**Parameters:**
+- `heritability_results`: The dict returned by `calculate_heritability_estimates`
+  (the `remove_low_h2=False` form, or the first element of the
+  `remove_low_h2=True` tuple)
+
+**Returns:**
+- DataFrame indexed by genotype (the union of every succeeded trait's `blup`
+  keys), one column per trait (excluding `__calculation_metadata__`), in the
+  input's trait order. `adjusted_mean = intercept + blup[genotype]` for each
+  succeeded trait/genotype cell. A trait whose model failed, used the
+  ANOVA-based or no-variance path, or was skipped gets an entire `NaN`
+  column — not omitted, not zero-filled. A genotype missing from one
+  succeeded trait's `blup` dict but present in another's gets a cell-level
+  `NaN` for that genotype/trait pair only. Never raises: a run-level
+  short-circuit dict (`{"error": "..."}`) returns an empty `pd.DataFrame()`.
+
+**Example:**
+```python
+heritability_results = calculate_heritability_estimates(df, trait_cols)
+blup_table = extract_blup_table(heritability_results)
+```
+
+---
+
 ## `pca` Module
 
 Principal Component Analysis for dimensionality reduction of root trait data.
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,6 +7,33 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- `extract_blup_table(heritability_results)` (#109): builds a genotype x trait
+  BLUP-adjusted-means `pd.DataFrame` from a `calculate_heritability_estimates` result —
+  `adjusted_mean = intercept + blup[genotype]` for each trait whose mixed model
+  succeeded, with a genuine `NaN` column (not dropped, not zero-filled) for any
+  trait that failed. Importable from `sleap_roots_analyze`. Tier 1 of the
+  cross-platform genotype-prediction program: BLUP-adjusted means are the predictor
+  substrate later tiers use to test whether one phenotyping platform's genotype
+  effects predict another's.
+- Serializable `BLUPResult` dataclass and `BLUPResult.from_blup_table(df, *,
+  intercepts=None)` adapter (#109): holds the genotype x trait adjusted-means matrix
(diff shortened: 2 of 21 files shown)


eberrigan (commented):
> **Verdict: APPROVE** (posted as comment — cannot approve your own PR)

## Review Summary

Five independent adversarial subagents (Code Quality, Testing Strategy, Statistical Rigor, Performance/Cross-Platform, Behavioural Correctness) reviewed this PR from scratch — none trusted the PR description's claims about the two previously-fixed bugs; each independently re-derived and re-verified them. **No BLOCKING issues found by any reviewer.** Both self-reported fixes (genotype identity in `08_blup_adjusted_means.csv`, and `BLUPResult.adjusted_means`'s shape invariant) were independently reproduced and confirmed genuinely fixed — not just claimed. Statistical correctness of the BLUP extraction itself (random-intercept-only model, `intercept + blup[g]`, NaN handling, non-mutation, reproducibility) was verified line-by-line by the statistical rigor and behavioural correctness passes. Code quality scored 9/10; TDD discipline verdict was PASS with real independent oracles (`df.groupby("geno")[trait].mean()`), not tautological self-comparison.

## Blocking Issues

None.

## Important Issues

- **Testing Strategy / Statistical Rigor (same gap, two reviewers independently flagged it):** the pipeline-level CSV test (`test_blup_csv_written_when_both_enabled`) only exercises a dataset where every trait succeeds, and only asserts genotype-row *set* equality — not full per-row value correctness, and not the NaN-column-survives-the-full-pipeline-and-CSV-round-trip path for a genuinely failing trait. The strong per-value/failed-trait proofs currently live only at the `extract_blup_table`/`BLUPResult` unit-test layer. Recommend adding one pipeline-level test with a trait engineered to fail its mixed-model fit, asserting the resulting CSV has an all-NaN column for it and correct per-genotype values for the succeeding traits.
- **Performance/Cross-Platform:** the `assert isinstance(heritability_results, dict)` added for mypy (commit `a8fa8a0`) is correct today (the call site hardcodes `remove_low_h2=False`), but `assert` is stripped under `python -O`/`PYTHONOPTIMIZE`. It's not a live risk now, but if a future refactor ever threads a variable through in place of the literal, this guard silently vanishes in optimized mode. Consider an explicit `if not isinstance(...): raise TypeError(...)` (or a comment flagging the assert is a static-analysis aid, not a runtime safety net) so this doesn't quietly bit-rot.

## Suggestions

- A `@overload` pair on `calculate_heritability_estimates` keyed on `Literal[True]`/`Literal[False]` for `remove_low_h2` would let mypy narrow all call sites (this one and two pre-existing ones) statically instead of asserting locally — real but pre-existing typing debt, worth a follow-up issue rather than blocking this PR.
- A debug-level log line when `generate_blup_table=True` + `calculate_heritability=False` silently no-ops (design.md Decision 5) would aid discoverability without reintroducing the warning-noise problem that decision explicitly rejected.
- Minor comment suggestions (documenting the `"error"`-key vs. real-trait-named-"error" distinction in `extract_blup_table`, vectorizing `extract_blup_table`'s per-trait loop if genotype/trait counts ever grow past current scale) — optional, not required.

All CI checks pass (lint, mypy baseline, tests on ubuntu/windows/macos, serialization/reproducibility/numerical-stability gates). No Copilot comments exist on this PR to reconcile.

---

PR #193, opened 2026-07-16
feat(#114): fixed_effects param for heritability mixed model (Tier 2 cross-platform prediction)

### LINKED ISSUE #114: [feature] Allow fixed effects in heritability mixed-model (e.g., experiment, wave, batch)
## Motivation

`calculate_heritability_estimates()` currently fits a fixed mixed model with genotype as the only effect:

```python
# src/sleap_roots_analyze/statistics.py:290-293
smf.mixedlm("value ~ 1", model_data, groups=model_data["genotype"])
```

i.e. `value ~ 1 + (1|genotype)`. There is no way to add fixed-effect covariates (experiment, wave, batch, scanner, etc.).

This is a problem for **multi-experiment GWAS panels** where genotypes are not perfectly balanced across experiments. Any systematic per-experiment shift in a trait (e.g., scanner calibration drift, cylinder placement, image pre-processing batch differences) gets absorbed into the genotype term and **inflates apparent H²**.

## Concrete example from a recent run

Mauricio Chiurazzi's combined alfalfa GWAS — 35 accessions across 6 Bloom experiments (combined for analysis):

- `Scanline First Ind P25` reports H² ≈ 0.83 (top of the heritability ranking)
- BUT per-experiment means of this trait shift ~1.5 scanline bins between Bloom experiments (Group3 mean = 7.32 vs Alfalfa_GWAS mean = 8.91) — a 25% shift across the trait's range
- Genotypes are not balanced across experiments (each Bloom experiment has its own subset of accessions), so the calibration drift is confounded with genotype
- Within-genotype variance > between-genotype variance for this trait family (naive ICC = 0.087–0.318) — but the mixed-model's H² lands at 0.80+ because experiment-level shifts get credited to genotype

We can't easily detect or correct for this without fitting `value ~ experiment + (1|genotype)` and comparing corrected vs uncorrected H².

See [talmolab/sleap-roots#172](https://github.com/talmolab/sleap-roots/issues/172) for the underlying scanner-calibration discussion.

## Proposed API

Add an optional `fixed_effects` parameter:

```python
def calculate_heritability_estimates(
    df: pd.DataFrame,
    trait_cols: List[str],
    genotype_col: str = "geno",
    replicate_col: str = "rep",
    fixed_effects: Optional[List[str]] = None,  # NEW
    ...
):
    ...
    if fixed_effects:
        formula = "value ~ " + " + ".join(fixed_effects)
    else:
        formula = "value ~ 1"
    model = smf.mixedlm(formula, model_data, groups=model_data["genotype"])
```

And expose it in the QC config schema:

```yaml
heritability:
  enabled: true
  threshold: 0.0
  fixed_effects: [experiment_name, wave_number]  # NEW
```

## Acceptance criteria

- `calculate_heritability_estimates()` accepts a `fixed_effects: Optional[List[str]] = None` parameter
- When provided, the mixed-model formula becomes `value ~ <fixed_effects> + (1|genotype)`
- The variance partition still pulls from `result.cov_re` (genetic) and `result.scale` (residual) — same H² formula post-fit
- New parameter exposed in `HeritabilityConfig` and surfaced in the QC config
- Documentation note that fixed effects should ONLY be metadata-style covariates that confound with genotype, not biological traits
- Test coverage: a synthetic batch-confounded dataset where uncorrected H² > corrected H², matching the pattern this issue describes

## Related issues

- [talmolab/sleap-roots#172](https://github.com/talmolab/sleap-roots/issues/172) — Scanline First Ind quantization / calibration drift docs
- #111, #112, #113 — other findings from the same alfalfa GWAS run

eberrigan:
## Summary

Tier 2 of the wheat EDPIE cross-platform genotype-prediction program (Tier 1: BLUP extraction, #189/#190). Adds a `fixed_effects` parameter to `calculate_heritability_estimates()` so the mixed model becomes `value ~ <fixed_effects> + (1|genotype)` instead of genotype-only, fixing a heritability-inflation bug where a batch/experiment confound gets absorbed into the genotype term when genotypes aren't balanced across batches (concrete case in #114: alfalfa GWAS trait reporting H² ≈ 0.83 despite naive within-genotype ICC of 0.087–0.318).

- `calculate_heritability_estimates(fixed_effects: Optional[List[str]] = None)` — every name is `C(...)`-wrapped unconditionally (always categorical). `fixed_effects=None` (default) is byte-for-byte identical to prior behavior.
- New `_marginal_intercept()` helper: when `fixed_effects` is set, `intercept` becomes an empirical, sample frequency-weighted value rather than the raw reference-level `fe_params["Intercept"]`, so BLUP-adjusted means aren't pinned to patsy's arbitrary baseline category.
- A captured `ConvergenceWarning` (checked by category, not message text) is treated as a fit failure — only when `fixed_effects` is set — since `statsmodels` does not always raise on a fixed effect confounded with genotype.
- `StatisticsConfig.fixed_effects` (default `None`) threads through `StatisticalAnalysisStep`.
- `fixed_effects` is independent of `replicate_col` — a block/replicate fixed effect (R_j) is expressed by naming that column directly, confirming Tier 1's tentative design note.

## Process notes

This proposal went through 4 rounds of adversarial review (`openspec/changes/add-heritability-fixed-effects/design.md`) before implementation, including empirical `statsmodels` simulation that caught and fixed a synthetic oracle fixture that was silently wrong-signed. A 5-agent pre-merge review then caught and fixed:
- A real bug: `remove_low_h2=True` + `fixed_effects` could silently drop the fixed-effect column from the filtered output.
- A latent robustness gap: the marginal-intercept coefficient-matching guard was a count check, not an identity check.
- A confusing error path when a `fixed_effects` name duplicates `genotype_col`/`replicate_col`.
- A CI-fragile test tolerance on a platform-sensitive characterization test.

Full rationale for every design decision and review finding is in `design.md`.

## Test plan

- [x] Full test suite: 2644 passed, 31 skipped (`uv run pytest tests/ -m "not integration"`)
- [x] `black --check` / `ruff check` clean on all changed files
- [x] `openspec validate add-heritability-fixed-effects --strict` passes
- [x] Core oracle verified empirically against real `statsmodels` across 10+ seeds (not eyeballed) — batch-confounded fixture shows uncorrected H² > corrected H² with 4x+ margin at the worst seed
- [x] Backward compatibility: `fixed_effects=None` traced and tested as byte-for-byte identical to pre-change behavior
- [x] New regression tests for all pre-merge-review findings

Closes #114

### DIFF
--- a/configs/templates/viz_template_no_images.yaml
+++ b/configs/templates/viz_template_no_images.yaml
@@ -35,6 +35,12 @@ statistics:
   calculate_anova: true
   calculate_heritability: true
   alpha: 0.05
+  # Optional metadata covariates (e.g. batch/experiment/block/scanner -- not
+  # biological traits) to add as fixed effects to the heritability mixed
+  # model (issue #114). Automatically excluded from the trait_cols scan, so
+  # listing a name here does not also require adding it to
+  # data.additional_exclude_cols.
+  # fixed_effects: ["experiment"]
 
 # PCA configuration
 pca:
--- a/configs/templates/viz_template_with_images.yaml
+++ b/configs/templates/viz_template_with_images.yaml
@@ -44,6 +44,13 @@ statistics:
   # Significance level for ANOVA
   alpha: 0.05
 
+  # Optional metadata covariates (e.g. batch/experiment/block/scanner -- not
+  # biological traits) to add as fixed effects to the heritability mixed
+  # model (issue #114). Automatically excluded from the trait_cols scan, so
+  # listing a name here does not also require adding it to
+  # data.additional_exclude_cols.
+  # fixed_effects: ["experiment"]
+
 # PCA configuration
 pca:
   # Number of PCs to retain (should match QC config)
--- a/docs/API.md
+++ b/docs/API.md
@@ -362,6 +362,21 @@ Calculate broad-sense heritability (H²) for traits using mixed models.
   experiment, wave, batch, scanner) to add as fixed effects, changing the
   model formula from `value ~ 1` to `value ~ C(fe_1) + C(fe_2) + ...`
   (issue #114). Default `None` reproduces pre-existing behavior exactly.
+  A duplicate name within `fixed_effects` (e.g. `["experiment",
+  "experiment"]`) is rejected with a structural `{"error": ...}`, same as a
+  missing or invalid column name. When a mixed-model fit succeeds with zero
+  `ConvergenceWarning`s but a genotype's observations are confined to a
+  single level of a fixed effect that has more than one level overall
+  (a confound `statsmodels` doesn't reliably warn about on its own), a
+  `UserWarning` is emitted describing the possible confound; the trait
+  still returns a normal successful result (this is diagnostic only, not a
+  failure classification).
+- `StatisticsConfig.fixed_effects` (the pipeline config equivalent, used by
+  `VizPipelineConfig`) must be `None` or a `list` of `str` -- a bare string
+  is rejected at config-construction time. Names listed there are also
+  automatically unioned into `VizPipelineConfig.data.additional_exclude_cols`,
+  so a fixed-effect column is never mistaken for a phenotypic trait by the
+  pipeline's upstream `trait_cols` scan.
 
 **Returns:**
 - If `remove_low_h2=False`: Dictionary with heritability results
--- a/docs/API.md
+++ b/docs/API.md
@@ -341,7 +341,8 @@ calculate_heritability_estimates(
     remove_low_h2: bool = False,
     h2_threshold: float = 0.3,
     barcode_col: str = "Barcode",
-    additional_exclude: Optional[List[str]] = None
+    additional_exclude: Optional[List[str]] = None,
+    fixed_effects: Optional[List[str]] = None
 ) -> Union[Dict, Tuple[Dict, pd.DataFrame, List[str], Dict]]
 ```
 
@@ -357,6 +358,10 @@ Calculate broad-sense heritability (H²) for traits using mixed models.
 - `h2_threshold`: Heritability threshold for filtering (default: 0.3)
 - `barcode_col`: Name of sample ID column for preservation
 - `additional_exclude`: Additional columns to exclude from filtering
+- `fixed_effects`: Optional list of metadata-style covariate columns (e.g.
+  experiment, wave, batch, scanner) to add as fixed effects, changing the
+  model formula from `value ~ 1` to `value ~ C(fe_1) + C(fe_2) + ...`
+  (issue #114). Default `None` reproduces pre-existing behavior exactly.
 
 **Returns:**
 - If `remove_low_h2=False`: Dictionary with heritability results
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -37,6 +37,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   pipeline's `StatisticalAnalysisStep` writes `08_blup_adjusted_means.csv` alongside
   `08_heritability_results.csv`. Only takes effect when `calculate_heritability` is
   also `True` — free once the model is fit.
+- `StatisticsConfig.fixed_effects` must now be `None` or a `list[str]` — a bare
+  string is rejected with `ValueError` at config-construction time instead of
+  silently degrading to a per-character `fixed_effects` list (PR #193 review).
+- A `UserWarning` is now emitted when a fixed effect is confounded with genotype
+  (every observation for some genotype confined to a single level of that fixed
+  effect), even when the fit converges cleanly with zero `ConvergenceWarning`s —
+  diagnostic only, does not reclassify the trait's result (PR #193 review).
+- A duplicate name within `fixed_effects` (e.g. `["experiment", "experiment"]`) is
+  now rejected with a structural `{"error": ...}`, matching the existing
+  missing-column/reused-name error shape, instead of an obscure `patsy` failure
+  (PR #193 review).
+
+### Fixed
+- `VizPipelineConfig` now automatically unions `statistics.fixed_effects` into
+  `data.additional_exclude_cols` at construction time, so a fixed-effect column
+  named outside the hardcoded metadata-substring list (e.g. `"block"`) is no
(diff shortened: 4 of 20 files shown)


eberrigan (commented):
> **Verdict: REQUEST_CHANGES** (posted as comment — cannot request changes on your own PR)

## Review Summary

Five subagents (Code Quality, Testing/TDD, Statistical Rigor, Performance/Cross-Platform, Behavioural Correctness) reviewed this independently. The core statistics work is rigorous — `_marginal_intercept`'s frequency-weighting math, the `fixed_effects=None` byte-for-byte-identical guarantee, and the `ConvergenceWarning`-as-failure heuristic all check out, backed by genuinely validated fixtures (empirically re-signed across 10+ seeds after design review caught a wrong-signed oracle). However, two reviewers *independently* reproduced the same crash in the new upfront validation, and a third flagged a real gap in how `fixed_effects` interacts with the rest of the pipeline (not just the direct API). That combination is enough to request changes before merge, even though the fix is small.

## Blocking Issues

**Uncaught `AttributeError` when a `fixed_effects` element is non-`str` and coincidentally matches a real column — contradicts this function's own documented contract.**

`statistics.py:445` — `invalid_fe_names = [fe for fe in fixed_effects if not fe.isidentifier()]` assumes every element is a `str`. Reproduced directly by two independent reviewers:

```python
df = pd.DataFrame({"trait": [1, 2, 3, 4], "geno": ["a", "a", "b", "b"], 5: [1, 1, 2, 2]})
calculate_heritability_estimates(df, ["trait"], replicate_col=None, fixed_effects=[5])
# AttributeError: 'int' object has no attribute 'isidentifier'
```

An int-labeled column is plausible for CSV-derived batch/wave/scanner codes, and this check runs *before* the per-trait `try/except`, so it crashes the whole call — and, via `StatisticalAnalysisStep.execute()` (no try/except around this call either), the whole pipeline step — despite the function's own docstring stating "Raises: Nothing propagates to the caller." Fix: `if not isinstance(fe, str) or not fe.isidentifier()`.

## Important Issues

1. **`fixed_effects` columns are never excluded from `trait_cols` upstream of the statistics step**, so a misconfigured pipeline silently treats a batch/metadata column as a phenotypic trait for every step before it. `trait_cols` is fixed once in `LoadDataStep` via `get_trait_columns(..., additional_exclude=config.data.additional_exclude_cols)`, which has no knowledge of `config.statistics.fixed_effects` and only excludes names matching a hardcoded substring list (`date`, `batch`, `experiment`, `scan_`, …) — an arbitrary name like `"block"` slips through. The existing `remove_low_h2=True` fix for this (unioning `fixed_effects` into `additional_exclude`) only covers direct API callers; `StatisticalAnalysisStep` always calls with `remove_low_h2=False`, so the pipeline itself never benefits from it. Nothing validates or documents that `statistics.fixed_effects` and `data.additional_exclude_cols` must be kept in sync.

2. **No config-schema validation for `StatisticsConfig.fixed_effects`.** Every comparable field (`heritability` threshold, `group_by` column existence, etc.) fails fast in `validate_qc_config()`/`validate_viz_config()`; `fixed_effects` doesn't. Concretely, `fixed_effects="experiment"` (a bare string instead of a list — dataclasses don't enforce this at runtime) isn't caught, and `required_cols.extend(fixed_effects)` iterates the string's *characters*, producing a misleading `{"error": "Missing required columns: ['e', 'x', 'p', ...]"}` that looks like a data problem rather than a config-type problem.

3. **No golden-template documentation.** None of `configs/templates/*.yaml` mention `fixed_effects`, even as a commented-out example, so there's no in-repo guidance steering a user toward also updating `additional_exclude_cols` (compounds issue #1).

4. `test_near_fully_confounded_fixed_effect_organic_behavior` (`test_statistics.py:1078-1139`) asserts an exact `convergence_warnings == []` outcome on a near-singular fit with no mocking, and its own docstring admits this is "platform/BLAS-sensitive" and "observed once (seed=0, this platform)." It passed locally on Windows; worth a follow-up check once CI's mac/ubuntu/windows `Tests` jobs (still pending at review time) all report back, since this is the first time it runs on that matrix.

5. The duplicate-name validation (`fixed_effects` colliding with `genotype_col`/`replicate_col`) is implemented and tested but was never added to `specs/statistics-api/spec.md`'s normative requirement text — a 1:1 spec-to-test mapping gap.

## Suggestions

- Duplicate entries *within* `fixed_effects` itself (e.g. `["experiment", "experiment"]`) aren't rejected — degrades to a cryptic `mixed_model_failed` error from patsy rather than the clean, purpose-built message the genotype/replicate collision gets. Consider extending the existing `reused_names` check.
- Zero test coverage for the single-level (zero-variance) fixed-effect case — verified correct by hand-tracing `_marginal_intercept`, but untested.
- Given the confirmed false-negative in the `ConvergenceWarning` heuristic (a fixed effect near-deterministically confounded with genotype can fit cleanly with zero warnings — reproduced in the test above), consider surfacing this to end users as an actual `UserWarning` when `fixed_effects` is used, not just docstring prose.
- `_marginal_intercept`'s identity check dedups observed levels via `str()`; a mixed-dtype object column where two distinct raw values stringify identically could theoretically under-count. Narrow and already-documented-as-unlikely given fixed effects are metadata-style CSV columns — not worth blocking on.
- Two of the pre-merge-review regression tests (`remove_low_h2` + `fixed_effects` exclusion, duplicate-name rejection) have no corresponding entry in `tasks.md`, unlike the rest of this PR's unusually thorough TDD trail.

## What's solid (no action needed)

- `fixed_effects=None` is verified byte-for-byte identical to the pre-existing code path — confirmed by tracing every branch.
- The `remove_low_h2=True` + `fixed_effects` fix is correctly implemented for direct API callers.
- `replicate_col`/`fixed_effects` collision is already checked in both directions (confirmed contrary to an initial concern — `fe == genotype_col or fe == replicate_col` correctly handles `replicate_col=None`).
- No new avoidable copies, no redundant model fits, no path/OS-specific code, no global-warnings-state leak — `catch_warnings(record=True)` is properly scoped.
- Statelessness confirmed: `.rename()`/`.dropna()` never mutate the caller's `df` in place.
- Fixture validation is unusually rigorous: `heritability_data_batch_confounded` reliably produces uncorrected H² > corrected H² by ≥0.05, re-verified after design review caught an earlier wrong-signed draft.

---

PR #208, opened 2026-07-29
fix(#80): propagate PCA-filtered trait_names to UMAP and static figures

### LINKED ISSUE #80: UMAP and downstream steps use unfiltered traits after PCA zero-variance filtering

egao28:
## Summary

Fixes #80. Follow-up to #74/#76.

`PCAAnalysisStep` correctly computes a zero-variance-filtered trait list (`pca_results["feature_names"]`) but never wrote it back into `metadata["trait_names"]`/`valid_trait_names`. Every other filtering step in the pipeline (`cleanup_traits`, `filter_heritability`, `remove_outliers`, `detect_outliers`) keeps those two keys in lock-step as the shared "traits still in play" contract — PCA was the one step that computed a filtered set and didn't write it back.

- `PCAAnalysisStep` now updates `trait_names`/`valid_trait_names` to the post-filter feature set, and adds `original_trait_names` to preserve the pre-filter list.
- `UMAPAnalysisStep` inherits the fix automatically (direct metadata spread from the PCA step result).
- `GenerateStaticFiguresStep` does **not** inherit it automatically: the `VizPipeline` orchestrator's `_run_generate_static_figures` sources its primary metadata from a separate DAG branch (`02_calculate_statistics -> 06_heritability_analysis -> 08_genotype_aggregation`) that never touches PCA. Its existing PCA-branch cherry-pick (`pca_results`/`top_features`/`n_pca_components`/`pca_explained_variance`) is extended to also copy `trait_names`/`original_trait_names` — fixing `pca_analysis.py` alone would have been a no-op for every static figure.
- This also fixes a latent bug in `create_pca_biplot`, which indexes its `trait_names` argument positionally against `pca_results["loadings"]` and would silently mislabel feature arrows whenever an excluded trait wasn't the last column of the original trait list (same bug class as #183, reached via the pipeline metadata path instead of a producer's own return dict).

Full investigation, severity analysis, and the design rationale (why two files need to change, and why `create_feature_contribution_plot` is *not* affected in current usage) are in `openspec/changes/fix-post-pca-trait-names-propagation/design.md`.

## Process

- OpenSpec proposal drafted, then critically reviewed by a 5-agent adversarial review (spec quality, code/architecture feasibility, GitHub issue alignment, TDD/testing strategy, data traceability + git workflow). All findings addressed: corrected an overstated claim about `create_feature_contribution_plot`, fixed an untestable "PCA disabled" scenario (no such config flag exists), added a missing regression scenario/task, fixed fixture-reuse and test-specificity gaps, and split the deep investigation out into `design.md`.
- TDD: tests written first against an **interleaved** zero-variance fixture (existing fixtures only have trailing constant columns, which can't distinguish "filtered by value" from "filtered by trailing-slice coincidence"), confirmed red, then the two-file fix applied, confirmed green.

## Test plan

- [x] New regression tests in `test_step_pca_analysis.py`, `test_step_umap_analysis.py`, `test_viz_pipeline_zero_variance.py` — all pass
- [x] Full existing suite for the three modified test files + `test_step_generate_static_figures.py`: 115/115 pass, no regressions
- [x] `ruff check src/sleap_roots_analyze` (CI's scope): clean
- [x] `black --check src/sleap_roots_analyze tests`: clean
- [x] mypy vs. frozen baseline: 0 new errors
- [ ] **Full ~1939-test suite**: not run locally — four attempts (background and foreground, with explicit long timeouts) failed to complete in this environment without producing usable output, which looks like a local tooling/environment limitation rather than a test failure, since the same approach reliably completes for smaller scoped runs. CI runs this exact suite (`pytest -m "not integration" tests/`) across Ubuntu/Windows/macOS and is the authoritative full-suite gate here — flagging this explicitly rather than claiming untested coverage.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -113,6 +113,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   (PR #193 review).
 
 ### Fixed
+- `PCAAnalysisStep` now updates `metadata["trait_names"]`/`valid_trait_names`
+  to the zero-variance-filtered feature set after PCA, instead of leaving
+  them as the pre-filter list while only recording
+  `excluded_zero_variance_traits`/`n_traits_after_filtering` alongside them
+  (#80). The pre-filter list is preserved under a new
+  `metadata["original_trait_names"]` key. Downstream, `UMAPAnalysisStep`
+  inherited the stale list directly (silently re-including excluded traits
+  in `feature_cols` and the logged `n_traits`), and the `VizPipeline`
+  orchestrator's `_run_generate_static_figures` cherry-picks `trait_names`/
+  `original_trait_names` from the PCA branch alongside `pca_results`/
+  `top_features` (previously omitted), so static figures now use the
+  corrected set too — this also fixes a latent bug in `create_pca_biplot`,
+  which indexes its `trait_names` argument positionally against
+  `pca_results["loadings"]` and would silently mislabel feature arrows when
+  an excluded trait wasn't the last column of the original trait list.
 - `VizPipelineConfig` now automatically unions `statistics.fixed_effects` into
   `data.additional_exclude_cols` at construction time, so a fixed-effect column
   named outside the hardcoded metadata-substring list (e.g. `"block"`) is no
--- a/openspec/changes/fix-post-pca-trait-names-propagation/design.md
+++ b/openspec/changes/fix-post-pca-trait-names-propagation/design.md
@@ -1,0 +1,134 @@
+## Context
+
+Issue #80 reports that `PCAAnalysisStep` filters zero-variance traits into
+`pca_results["feature_names"]` but never writes that filtered list back into
+pipeline metadata, so downstream steps reading `metadata["trait_names"]` get
+the pre-PCA list. The issue's own "Proposed Fix" section suggests a single
+change: update `pca_analysis.py`'s metadata dict. This document records why
+that alone is insufficient, and what the actual fix requires.
+
+`VizPipeline` (`viz_pipeline.py`) runs 12 steps as a NetworkX DAG, not a
+strict linear chain. Two steps read `trait_names` downstream of PCA, and
+they reach it through different paths:
+
+```
+02_calculate_statistics ─┬─→ 03_pca_analysis ─┬─→ 04_umap_analysis
+                          │                    └─→ 05_cluster_analysis
+                          └─→ 06_heritability_analysis ─→ 08_genotype_aggregation ─→ 09_generate_static_figures
+```
+
+`04_umap_analysis` receives `prev_result` **directly** from `03_pca_analysis`
+(`viz_pipeline.py:333-344`), and `UMAPAnalysisStep.execute` spreads
+`**prev_result.metadata` verbatim (`umap_analysis.py:161-165`). So once
+`PCAAnalysisStep` writes the corrected `trait_names`, UMAP inherits it for
+free — no change needed in `viz_pipeline.py` for that path.
+
+`09_generate_static_figures` does **not** receive `prev_result` from PCA at
+all. Its primary metadata source is `08_genotype_aggregation`
+(`viz_pipeline.py:414-465`), which is a pure passthrough
+(`genotype_aggregation.py:31`) of `06_heritability_analysis`'s metadata,
+itself forked from `02_calculate_statistics` — a branch that never touches
+PCA. The orchestrator already knows it needs specific values from the PCA
+branch and hand-picks them into `combined_metadata`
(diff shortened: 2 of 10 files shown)


eberrigan (approved):
## Review Summary

Clean, well-scoped bug fix for #80. `PCAAnalysisStep` now writes its already-computed zero-variance-filtered trait list back into `metadata["trait_names"]`/`valid_trait_names"` (matching the "traits still in play" contract every other filtering step maintains), adds a traceable `original_trait_names` key, and `viz_pipeline.py`'s static-figures orchestrator is correctly extended to cherry-pick the same keys across its asymmetric DAG branch. Five independent review passes (code quality/architecture, testing/TDD, statistical rigor/reproducibility, performance/cross-platform, behavioural correctness) found **no blocking issues** — CI is green on Ubuntu/Windows/macOS, and every claim in the PR description was traced against actual source and confirmed accurate, including the secondary `create_pca_biplot` positional-mislabeling fix and the (correct) decision not to touch `create_feature_contribution_plot` or `generate_summary_viz.py`. TDD discipline is well-evidenced (test-only commit precedes the fix commit; `tasks.md` documents specific pre/post-fix failure numbers).

## Notes (non-blocking)

- **Test coverage is one level removed from the claimed `create_pca_biplot` fix**: the new tests assert `GenerateStaticFiguresStep`'s `trait_names` metadata is correct, but nothing directly calls `create_pca_biplot` and asserts arrow-label identity/order against `loadings.index`. The underlying logic was manually verified correct by two independent reviewers, but a narrow unit test on `create_pca_biplot` itself would close this gap and guard against future regressions in `generate_static_figures.py`'s pass-through.
- The new spec scenario "trait distribution plots SHALL NOT be generated for excluded zero-variance traits" (`specs/visualization-pipeline/spec.md`) has no corresponding test.
- Minor latent robustness gaps, both currently unreachable and not live bugs: (1) `metadata["trait_names"]`, `metadata["valid_trait_names"]`, and `pca_results_serializable["feature_names"]` all alias the same `feature_names` list object rather than independent copies — no current code mutates it in place, but a future in-place mutation on one key would silently corrupt the others (the PR already does the defensive `list(...)` copy for `original_trait_names`; worth the same treatment here for consistency). (2) `generate_static_figures.py`'s `metadata.get("trait_names", metadata.get("valid_trait_names", []))` fallback wouldn't trigger if `trait_names` were present-but-`None` — unreachable today since `pca_analysis.py` always sets a non-`None` list, but an `or`-chain (mirroring the pattern already used in `pca_analysis.py`/`umap_analysis.py`) would be more defensive.
- `create_umap_colored_by_top_traits` (`visualization.py`) has the same positional-indexing pattern as `create_pca_biplot` and also benefits from this fix, but isn't mentioned in the PR/changelog — worth a one-line note for completeness.

None of the above block merge; they're good fast-follow or nice-to-have items.

---

PR #214, opened 2026-07-30
fix(#202): honor feature_selection in PCA biplot and contribution plot

eberrigan:
## Summary

Fixes #202. `pca.feature_selection_strategy` was silently not controlling feature selection in two visualization functions:

- **`create_pca_biplot`**: had `if`/`elif` branches for `extreme`/`top_absolute`/`top_contribution`/`vector_length` but none for `top_variance`, so it silently fell through to `vector_length` — a materially different, differently-weighted selection criterion. Now has an explicit `top_variance` branch, and raises `ValueError` for genuinely unrecognized values instead of silently substituting `vector_length`. When `feature_selection="top_variance"`, `pc_indices=None` is passed (not the 2 displayed PCs), since `select_top_features_from_pca`'s `top_variance` method ranks across all retained PCs and ignores `pc_indices` entirely — mirrors the existing workaround in `create_umap_colored_by_top_traits`.
- **`create_feature_contribution_plot`**: accepted a `feature_selection` parameter that was never referenced anywhere in the function body — every code path always ranked by total variance contribution regardless of what was passed. **Removes** the parameter (breaking, narrow) rather than wiring it up: the chart's own title asserts the displayed traits are the top contributors, so a non-contribution selection criterion would misdescribe the chart, matching the parameter-free `create_feature_contribution_heatmap` precedent. Its on-the-fly ranking branch now delegates to `select_top_features_from_pca(method="top_variance")` instead of duplicating the formula inline.

Verified backward-compatible with the real downstream consumer (`Salk-Harnessing-Plants-Initiative/bloom`, `sleap-roots-analyze>=0.1.0a5`): its `pca_analysis` MCP tool calls both functions but never passes `feature_selection` to either.

Followed the repo's spec-driven + TDD workflow: OpenSpec proposal (`openspec/changes/fix-pca-feature-selection-wiring/`) → 4-agent proposal review → TDD red/green implementation → 5-agent pre-PR self-review → this PR.

## Test plan

- [x] New regression tests: `top_variance` selects the same features as a direct `select_top_features_from_pca(method="top_variance")` call (for both `create_pca_biplot` and `create_feature_contribution_plot`'s on-the-fly branch)
- [x] `create_pca_biplot` raises `ValueError` for an unrecognized `feature_selection`
- [x] Parametrized regression guard confirms the four pre-existing methods (`vector_length`/`extreme`/`top_absolute`/`top_contribution`) still dispatch correctly (previously untested)
- [x] `create_feature_contribution_plot`'s signature no longer accepts `feature_selection`, and the `GenerateStaticFiguresStep` call site no longer passes it (real step-execution test, not just a signature check)
- [x] Full local test suite: 2875 passed, 0 failed
- [x] `black --check`, `ruff check` (src), `mypy` vs. frozen baseline (0 new errors) all clean
- [x] `openspec validate fix-pca-feature-selection-wiring --strict` passes
- [x] Synthetic before/after visual sanity check on real trait data (`tests/data/Turface_all_traits_2024.csv`) confirmed the fix changes which features are plotted, as expected

### DIFF
--- a/.mypy-baseline.txt
+++ b/.mypy-baseline.txt
@@ -265,7 +265,6 @@ src/sleap_roots_analyze/visualization.py:0: error: Argument 1 to "len" has incom
 src/sleap_roots_analyze/visualization.py:0: error: Item "None" of "Any | None" has no attribute "head"  [union-attr]
 src/sleap_roots_analyze/visualization.py:0: error: Argument 1 to "len" has incompatible type "Any | None"; expected "Sized"  [arg-type]
 src/sleap_roots_analyze/visualization.py:0: error: Incompatible types in assignment (expression has type "signedinteger[_32Bit | _64Bit]", variable has type "int | None")  [assignment]
-src/sleap_roots_analyze/visualization.py:0: error: Incompatible types in assignment (expression has type "ndarray[tuple[Any, ...], dtype[signedinteger[_32Bit | _64Bit]]]", variable has type "list[Any]")  [assignment]
 src/sleap_roots_analyze/visualization.py:0: error: Module has no attribute "tab20"  [attr-defined]
 src/sleap_roots_analyze/visualization.py:0: error: Argument 3 to "linspace" has incompatible type "int | None"; expected "SupportsIndex"  [arg-type]
 src/sleap_roots_analyze/visualization.py:0: error: Argument 1 to "range" has incompatible type "int | None"; expected "SupportsIndex"  [arg-type]
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -113,6 +113,28 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   (PR #193 review).
 
 ### Fixed
+- `create_pca_biplot` now honors `feature_selection="top_variance"` (#202): the
+  `feature_selection`→`method` mapping had `elif` branches for `extreme`/
+  `top_absolute`/`top_contribution`/`vector_length` but none for `top_variance`,
+  so it silently fell into `else: method = "vector_length"` — a materially
+  different selection criterion (eigenvalue-weighted variance contribution vs.
+  unweighted Euclidean norm in the PC plane). An explicit `top_variance` branch
+  now passes `pc_indices=None` (rather than the biplot's 2 displayed PCs) since
+  `select_top_features_from_pca`'s `top_variance` method ranks across all
+  retained PCs and ignores `pc_indices` entirely. Any other `feature_selection`
+  value now raises `ValueError` instead of silently substituting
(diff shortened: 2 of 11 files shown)


eberrigan (commented):
> **Verdict: COMMENT (round 2)** — second independent 5-agent adversarial review pass, post-CI-fix and post-round-1-fixes.

## Review Summary

A second round of Code Quality, Testing, Statistical Rigor, Performance/Cross-Platform, and Behavioural Correctness reviews was run against the accumulated diff (now including the mypy-baseline sync commit). No BLOCKING or IMPORTANT issues were found. Three of the five reviewers independently converged on the same architectural suggestion, which has been applied.

## Findings & Resolution

- **Code Quality (9/10)**: confirmed the mypy-baseline sync is a genuine, causally-verified type-safety improvement (reverted the code locally, watched the baseline error reappear verbatim, restored it). Flagged that `create_pca_biplot`'s `feature_selection`→`method` mapping had become a pure identity `elif` chain duplicating validation `select_top_features_from_pca` already performs — drift risk if a 6th method is ever added without updating both places.
- **Testing (PASS)**: independently re-verified the round-1-added `test_create_feature_contribution_plot_on_the_fly_matches_direct_call` is correct (traced the branch-forcing condition and the reversal direction by hand, not just re-reading the comment) and re-ran the full affected suite (272 passed). Found one small, pre-existing, out-of-scope gap: no test directly asserts `select_top_features_from_pca(method="top_variance")` ignores a non-default `pc_indices` — filed as a suggested follow-up, not blocking this PR.
- **Statistical Rigor**: independently confirmed the `ValueError` change doesn't crash a `run-all` batch (each config runs as its own subprocess; a typo'd config now fails fast for *that* config instead of silently mis-plotting) and confirmed the mypy fix has zero runtime/ordering effect, only a type-annotation improvement. Echoed the same hardcoded-list-drift suggestion.
- **Performance/Cross-Platform**: nothing new; confirmed the `.mypy-baseline.txt` change and new test are both inert w.r.t. performance/platform.
- **Behavioural Correctness**: traced every edge case for the `ValueError` branch (`None`, `""`, wrong case, whitespace, non-string) — all correctly raise. No Copilot comments exist yet to triage.

## Action taken

Applied the repeatedly-flagged suggestion: added `pca.VALID_SELECTION_METHODS` as a single source of truth, used by both `select_top_features_from_pca`'s own error message and `create_pca_biplot`'s validation, replacing the redundant `elif` chain with a plain membership check. Verified: full affected-file suite (389 passed), `black`/`ruff` clean, `mypy-baseline` clean (0 new/fixed), `openspec validate --strict` clean. Pushed as `5026114`.

---
*Review by Claude Code subagent team (Code Quality · Testing · Statistical Rigor · Performance/Memory · Behavioural Correctness) — round 2*

PR #215, opened 2026-07-31
fix(#161): type files_generated as List[Path] in ReduceTraitRedundancyStep

egao28:
## Summary

First concrete step of #161 (CI: ratchet mypy strictness and pay the baseline down toward
zero). `ReduceTraitRedundancyStep` still had `files_generated` type annotations that
disagreed with its actual `Path` runtime values — #157/#159 removed the runtime
`str(path)` pre-stringification in this file but never fixed the surrounding type hints,
so `mypy-baseline sync` was a no-op despite the `cli-pipeline` spec's "Provenance Path
Serialization Is Centralized" invariant (`files_generated` SHALL be `List[Path]`) already
holding at runtime.

- Annotate `execute()`'s and `_cluster_experiment()`'s `files_generated` locals as
  `List[Path]`, and correct `_cluster_experiment`'s return-type annotation from `List[str]`
  to `List[Path]` for the files-list element.
- No behavior change — annotations only.
- Adds `isinstance(Path)` assertions to the two existing clustering-path tests in
  `tests/test_trait_redundancy.py`, since every prior assertion went through `str(f)` and
  would pass identically for `str` or `Path`.
- `.mypy-baseline.txt` shrinks 374 → 371 (3 lines: `var-annotated`, `arg-type`, and a
  `return-value` error the `Tuple` fix also resolves — the last one wasn't in the original
  proposal draft; caught during a 5-subagent OpenSpec review, see below).
- OpenSpec change: `openspec/changes/fix-files-generated-path-typing/` (extends the
  `cli-pipeline` spec's path-serialization requirement with a new scenario covering static
  annotation drift, not just runtime `str(path)` misuse).

**Out of scope** (explicitly, not silently): `generate_static_figures.py` has the same
`files = []` annotation-drift shape in two spots — left for its own follow-up PR to keep
this diff to one producer, per #161's own "each its own small PR" preference. The rest of
#161's checklist (pandas-stubs/types-*, `disallow_incomplete_defs`, `disallow_any_*`, the
mypy crash-guard in companion issue #160) is also untouched. **#161 stays open.**

## Review process

Ran this repo's `openspec-review`-style 5-subagent adversarial review before implementing.
It caught that the initial proposal draft had its own mypy accounting wrong: baseline is
374 lines (not 375), and the fix removes 3 lines (not 2) — an undocumented `return-value`
error at the corrected `Tuple` return type. It also caught that `generate_static_figures.py`
has a matching gap, which the initial draft's phrasing implied didn't exist. Both are
corrected in the proposal and reflected in the numbers above.

## Test plan

- [x] `uv run pytest tests/test_trait_redundancy.py -v` — 34/34 pass
- [x] `uv run pytest tests/test_predict_cross_platform_pipeline.py tests/test_step_calculate_trait_enrichment.py -q` — no regressions (49 passed, 1 pre-existing skip)
- [x] `uv run mypy src/sleap_roots_analyze | uv run mypy-baseline filter --baseline-path .mypy-baseline.txt` — 0 new errors, 3 fixed
- [x] `uv run mypy src/sleap_roots_analyze | uv run mypy-baseline sync --baseline-path .mypy-baseline.txt` — baseline shrinks by exactly 3 lines, no unrelated changes
- [x] `uv run ruff check src/sleap_roots_analyze` — clean
- [x] `uv run black --check src/sleap_roots_analyze tests` — clean
- [x] `openspec validate fix-files-generated-path-typing --strict` — valid


eberrigan (commented):
> **Verdict: COMMENT** — no blocking issues; two minor non-blocking items noted (test parity gap, confirm clean Windows CI re-run before merge)

## Review Summary

This is a narrowly scoped, well-verified type-annotation fix in `ReduceTraitRedundancyStep`. Five independent reviewers (code quality, testing, statistical rigor, performance/cross-platform, behavioural correctness) traced every code path that populates `files_generated` and confirmed the PR's central claim holds: this is genuinely annotation-only, with zero change to clustering logic, statistics, metadata flow, output schema, or runtime behavior. One reviewer re-ran black/ruff/mypy-baseline/pytest live against the branch in a scratch worktree and got results matching the PR description exactly (34/34 tests, 0 new mypy errors, baseline 374→371).

No blocking issues. Two minor items worth addressing before or shortly after merge.

## Blocking Issues

None.

## Important Issues

1. **Missing test parity**: `test_clustering_only_exp1` (`tests/test_trait_redundancy.py`) exercises the same `_cluster_experiment()` path as the two tests that received the new `assert all(isinstance(f, Path) for f in result.files_generated)` regression check, but wasn't given the same assertion. Given the PR's own stated rationale for adding these checks ("every current assertion goes through `str(f)`, which can't distinguish `Path` from `str`"), that rationale applies here too — looks like an oversight rather than a deliberate exclusion.
2. **CI**: the Windows job on this PR's only workflow run (30660513953) is marked cancelled (job conclusion `cancelled`, `gh pr checks` shows `fail`). The underlying `pytest` step itself completed successfully before the cancellation (2873 passed, 37 skipped, in 28m30s against a 30-minute job timeout), so this looks like an infra/timeout issue rather than a regression from this diff — but since this PR's subject is path-serialization correctness, please get a clean green re-run on Windows before merging rather than resting on a cancelled signal. (Separately, worth opening a follow-up issue: the Windows `tests` job appears to run persistently close to its 30-minute budget, independent of this PR — an infra concern, not something this PR needs to fix.)

## Suggestions

- Two new test comment blocks are duplicated verbatim across the two touched tests — fine at this scale, not worth abstracting.
- `assert all(isinstance(f, Path) for f in result.files_generated)` won't identify which element fails if it ever regresses; a `for f in ...: assert isinstance(f, Path), f` would localize failures better.
- No isinstance assertion covers the `trait_reduction_method == "none"` passthrough branch — low risk since it's a trivially empty list, but would complete coverage across all three branches (none / single-experiment / both).
- Worth noting for future reference: the new runtime assertions are a second, narrower layer on top of the pre-existing `tests/test_no_path_prestringify.py` AST guard, which already statically forbids `str(path)` from reaching `files_generated` across every step file. That's *why* this change carries essentially zero regression risk — it's not a hopeful assumption.

---

eberrigan (approved):

PR #218, opened 2026-07-31
CI: mypy crash-guard on exit code (#160)

egao28:
## Summary

- Adds a crash-guard to the `type-check` CI job's mypy step: captures mypy's own exit code via bash `PIPESTATUS` and fails CI immediately on anything outside `{0, 1}` (mypy's documented clean/errors-found convention), regardless of what `mypy-baseline filter` reports.
- Closes a known, unticketed hole flagged in #158's `design.md` ("Crash-safety becomes a hole at zero debt"): today the safety net only holds because `.mypy-baseline.txt` is non-empty (298 errors); once it's paid down to zero, a silent mypy crash (bad config, broken import, internal error) would look identical to a clean run and CI would false-green.
- Documents the guard in `docs/CONTRIBUTING.md`'s existing mypy-ratchet section.

Closes #160.

## OpenSpec

- Proposal: `openspec/changes/add-mypy-crash-guard/` (MODIFIED `Type-Check Gate` requirement in `developer-tooling` spec).
- Went through this repo's 5-agent OpenSpec review before implementation. The review caught a real, hands-on-verified BLOCKING bug in the original design: reading `PIPESTATUS[0]`/`PIPESTATUS[1]` as two separate statements loses the second value (bash resets `PIPESTATUS` after any intervening command, even a bare assignment) — the naive literal implementation would have broken CI on every currently-passing PR. Fixed by capturing the whole array atomically in one statement (`status=("${PIPESTATUS[@]}")`) before reading either element.
- `openspec validate add-mypy-crash-guard --strict` passes.

## Test plan

Verified locally against the exact `run:` block (not just reasoned about), covering every case in `tasks.md` §2:
- [x] Clean tree (mypy exit `0`) → job passes (`step_exit=0`)
- [x] Real baseline, real tree (mypy exit `1`, all errors covered) → job passes (`step_exit=0`), unchanged from today
- [x] New untyped def introduced (mypy exit `1`, new error) → job fails (`step_exit=1`), unchanged from today
- [x] Simulated fatal mypy exit (bad `--config-file`, exit `2`) against an **empty** scratch baseline → job fails via the new crash-guard (`step_exit=1`); confirmed this is exactly the false-green case that would have passed (`filter` alone reports `new: 0`) without this change
- [x] Simulated fatal mypy exit against the **real**, non-empty baseline → job still fails (guard fires; would also have failed via `filter`'s own logic)
- [ ] Confirm the real `type-check` job on this PR reflects the same result, and other jobs (lint, reproducibility, numerical-stability, serialization) are unaffected

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,10 +65,26 @@ jobs:
         shell: bash
         # pipefail is ON by default in Actions; mypy always exits non-zero while the
         # baseline holds errors, so disable it and let `filter`'s exit code (non-zero only
-        # on NEW errors) decide pass/fail.
+        # on NEW errors) decide pass/fail. That safety only holds while the baseline is
+        # non-empty though: once it's paid down to zero, a mypy crash that emits nothing
+        # looks identical to a clean run. Guard against that (issue #160) by capturing
+        # mypy's own exit code via PIPESTATUS and failing on anything outside {0, 1}
+        # (0=clean, 1=errors reported — mypy's documented convention; anything else is a
+        # fatal/usage/internal error or a signal kill). PIPESTATUS must be copied into a
+        # plain array in the very next statement — reading its elements as two separate
+        # statements loses the second one, since bash resets PIPESTATUS after every
+        # subsequent simple command.
         run: |
           set +o pipefail
           uv run mypy src/sleap_roots_analyze | uv run mypy-baseline filter --baseline-path .mypy-baseline.txt
+          status=("${PIPESTATUS[@]}")
+          mypy_status="${status[0]}"
+          filter_status="${status[1]}"
+          if [[ "$mypy_status" != "0" && "$mypy_status" != "1" ]]; then
+            echo "::error::mypy exited with fatal status $mypy_status (expected 0=clean or 1=errors-found); treating as a crash, not a baseline diff."
+            exit 1
+          fi
+          exit "$filter_status"
 
   reproducibility-gates:
     name: Reproducibility gates (determinism)
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -130,6 +130,14 @@ summary glyph can raise `UnicodeEncodeError` locally — prefix the command with
 
 Strictness starts lenient (one knob today) and is tightened over time in small follow-up PRs.
 
+CI also guards against mypy itself crashing (issue #160): a fatal/internal-error exit (anything
+other than mypy's documented `0`=clean or `1`=errors-reported) fails the job immediately, with a
+distinct `mypy exited with fatal status ...` annotation, regardless of what `mypy-baseline filter`
+reports. This matters most once the baseline reaches zero — without it, a silent crash would look
+identical to a clean run. The local commands above intentionally don't carry this guard: a local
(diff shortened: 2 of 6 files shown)


eberrigan (approved):

eberrigan (approved):
## Review Summary

Small, well-scoped CI change that closes a real, previously-flagged hole (#160, from #158's design.md): once `.mypy-baseline.txt` reaches zero errors, a silently-crashing mypy would look identical to a clean run to `mypy-baseline filter`, false-greening CI. The `PIPESTATUS`-based guard genuinely closes this. All 5 review lenses (code quality, testing/TDD, statistical rigor, performance/cross-platform, behavioral correctness) converged: no blocking issues. Statistical rigor and performance/memory checks are N/A — this PR touches no `src/sleap_roots_analyze` code, only `.github/workflows/ci.yml`, `docs/CONTRIBUTING.md`, and the OpenSpec proposal. The `PIPESTATUS`-atomic-capture bug this PR's own review process caught (splitting the read across two statements silently loses the second value) was independently reproduced and confirmed real.

## Important Issues (non-blocking)

1. **`tasks.md` §2.5's recorded verification number doesn't reflect true GitHub Actions shell semantics.** Reproduced independently by three review lenses under `bash -e -o pipefail` (GH Actions' actual invocation, confirmed from this PR's own job log: `/usr/bin/bash --noprofile --norc -e -o pipefail {0}`): for `mypy_status=2, filter_status=100`, real CI would exit `100`, not the documented `1`. Cause: with `pipefail` disabled, the pipe's reported status is `filter`'s; `errexit` (still on) aborts the script *at the pipe line itself* whenever that status is nonzero — before `status=("${PIPESTATUS[@]}")` ever runs. Practically inconsequential (both values are nonzero, so the job still correctly fails), but it means the guard's own `if`/`::error::` diagnostic only ever fires when `filter_status==0` — which is exactly the primary false-green case it exists for, so the fix itself is sound. Suggest correcting the recorded number in `tasks.md` (or noting the errexit short-circuit explicitly) so it doesn't mislead a future reader debugging this step.
2. **`design.md`'s Risks section undersells the scope of the same errexit interaction.** It frames the diagnostic-suppression trade-off as specific to "a crash that also makes `filter` exit non-zero on garbled input," but it actually applies to *any* nonzero `filter_status` — including the routine "new untyped def" case, and `mypy-baseline filter` crashing on its own independent of mypy. Worth a one-line addendum for future readers.
3. **The crash-guard's failure branch has zero automated-CI coverage** — every case where the guard actually fires has only been exercised via manual local verification (following this repo's existing #158 precedent, so not a new gap). Since this guard exists specifically to catch a *future* silent regression, consider extracting it into a small versioned script (e.g. `scripts/ci/mypy_guard.sh`) that a follow-up could cover with a trivial subprocess-based test, so the guard can't itself silently regress.

## Suggestions

- Minor: `design.md`/`proposal.md` cite "298 baseline errors"; the actual CI log on this PR shows "Found 295 errors... unresolved: 371" — a stale figure if anyone cross-references it later, not a defect in this PR.
- The trailing `exit "$filter_status"` is only reachable when `filter_status == 0` (per the errexit analysis above) — harmless and still the right defensive choice, just noting it's effectively always `exit 0` in practice today.

---

Issue #64, opened 2026-02-17
Cleanup: remove unused feature_selection parameter from create_feature_contribution_plot()


eberrigan:
## Summary

`create_feature_contribution_plot()` accepts a `feature_selection` parameter (sourced from `config.pca.feature_selection_strategy`) but silently ignores it. This is dead code that creates a misleading interface.

## Location

- **Call site**: `src/sleap_roots_analyze/pipeline/steps/generate_static_figures.py` ~line 414
- **Function**: `src/sleap_roots_analyze/visualization.py` → `create_feature_contribution_plot()`

## How the chart actually works (correctly)

The chart has two independent controls that are both working correctly:

1. **`variance_threshold`** (inherited from `pca.n_components`, e.g. 0.95) — determines which **PCs** appear as stacked bar segments. Ensures the chart reflects the same PCs used in scatter plots.
2. **`top_n`** (`config.static_viz.feature_contribution_top_n`, default 20) — limits how many **trait rows** are displayed. With ~1000 traits, this is needed for readability.

Traits are always ranked by total variance contribution across the included PCs. This is the correct behavior regardless of `feature_selection_strategy`.

## The problem

The `feature_selection` parameter is passed in and accepted by the function signature but never used in the selection logic:

```python
# generate_static_figures.py — passed in
contrib_fig = create_feature_contribution_plot(
    ...
    feature_selection=config.pca.feature_selection_strategy,  # ← ignored
)

# visualization.py — always top variance regardless
sorted_indices = np.argsort(total_contributions)[::-1]
top_indices = sorted_indices[:actual_top_n]
```

## Fix

Remove `feature_selection` from `create_feature_contribution_plot()`'s signature and from the call site. Add a docstring clarifying that traits are always ranked by total variance contribution.

## Impact

- No incorrect results — behavior is already correct
- Misleading to users who set `strategy: "extreme"` expecting the bar chart to change
- Dead code

## Priority

Low — cleanup. Fix alongside any future PCA visualization refactor.

PR #220, opened 2026-08-03
Fix #206: variance-driven PCA top-feature selection (+ #203 pc_indices scoping)

eberrigan:
## Summary

Fixes #206 and resolves the open "decide" checkbox in #203.

- **#203**: `PCAAnalysisStep` now always passes `pc_indices=list(range(n_components))` to `select_top_features_from_pca()`, instead of silently relying on that function's `[0, 1]` default. Runs with `n_components >= 3` previously only ever considered PC1/PC2 for `"extreme"`/`"top_absolute"`/`"top_contribution"`.
- **#206 Part 1**: `feature_selection_strategy: "extreme"` no longer takes an arbitrary count — it always selects exactly 1 most-positive-loading and 1 most-negative-loading trait per retained PC. `pca.n_top_features` is not read for this method (a `logger.info()` notice fires if it's set alongside `"extreme"`), and the 28 active configs pairing the two have had the now-meaningless line removed.
- **#206 Part 2**: `pca.n_top_features` changes type from `int` to `float` and gains a variance-threshold overload for `"top_variance"`, mirroring `pca.n_components`'s existing `< 1` = ratio / `>= 1` = count convention. A new `select_n_features_by_variance()` helper resolves the threshold to a concrete count.
- New config validation (`validate_qc_config()`/`validate_viz_config()`) rejects `n_top_features < 1` for `"top_absolute"`/`"top_contribution"` and rejects a non-whole-number `n_top_features >= 1` for every strategy except `"extreme"`.

Full design rationale (including the "ignore vs. validate" decision for `"extreme"`, why the new helper is a separate function, and why `"vector_length"` is intentionally excluded) is in the OpenSpec proposal: `openspec/changes/fix-pca-top-feature-selection/`. The proposal went through two rounds of adversarial multi-agent review before implementation.

## Test plan

- [x] New unit tests for `select_n_features_by_variance()` (`tests/test_pca.py`)
- [x] New regression tests for `PCAAnalysisStep`'s pc_indices scoping, extreme's n_top_features-ignoring + logging, and top_variance's threshold resolution (`tests/test_step_pca_analysis.py`)
- [x] New config-validation tests for the fractional/non-whole-number rejection (`tests/test_pipeline_config.py`, `tests/test_viz_pipeline_config.py`)
- [x] New parametrized regression test loading + validating all 28 edited configs (`tests/test_viz_pipeline_config.py`)
- [x] Full suite: `uv run pytest --cov=src/sleap_roots_analyze --cov-report=xml --durations=20 -m "not integration" tests/` — 2950 passed, 37 skipped, 0 failed
- [x] `uv run black --check` / `uv run ruff check` clean on all touched files
- [x] `openspec validate fix-pca-top-feature-selection --strict` passes

## Follow-up after merge

Per the proposal, #203 will be commented on with a cross-link to this PR and closed as resolved (its "decide" checkbox is answered here), matching the pattern used for #64/#68.

### DIFF
--- a/.mypy-baseline.txt
+++ b/.mypy-baseline.txt
@@ -4,7 +4,6 @@ src/sleap_roots_analyze/pipeline_runner.py:0: error: Incompatible types in assig
 src/sleap_roots_analyze/pipeline_runner.py:0: error: Argument 1 to "append" of "list" has incompatible type "tuple[Any, None]"; expected "tuple[Any, float]"  [arg-type]
 src/sleap_roots_analyze/pipeline_runner.py:0: error: Argument 1 to "_format_removed_traits_section" of "PipelineRunner" has incompatible type "list[tuple[str, list[tuple[Any, float]]]]"; expected "list[tuple[str, list[tuple[str, float | None]]]]"  [arg-type]
 src/sleap_roots_analyze/pipeline_runner.py:0: error: Need type annotation for "lines" (hint: "lines: list[<type>] = ...")  [var-annotated]
-src/sleap_roots_analyze/pca.py:0: error: Incompatible types in assignment (expression has type "signedinteger[_32Bit | _64Bit]", variable has type "int | None")  [assignment]
 src/sleap_roots_analyze/data_utils.py:0: error: Function is missing a type annotation  [no-untyped-def]
 src/sleap_roots_analyze/data_utils.py:0: error: Function is missing a type annotation  [no-untyped-def]
 src/sleap_roots_analyze/data_utils.py:0: error: Function is missing a type annotation  [no-untyped-def]
--- a/configs/active/viz/alfalfa_gwas_groups_1_to_6_combined.yaml
+++ b/configs/active/viz/alfalfa_gwas_groups_1_to_6_combined.yaml
@@ -28,7 +28,6 @@ pca:
   n_components: 0.75
   standardize: true
   feature_selection_strategy: "extreme"
-  n_top_features: 5
 
 umap:
   enabled: true
--- a/configs/active/viz/alfalfa_gwas_groups_1_to_6_combined_no_root_widths.yaml
+++ b/configs/active/viz/alfalfa_gwas_groups_1_to_6_combined_no_root_widths.yaml
@@ -22,7 +22,6 @@ pca:
   n_components: 0.75
   standardize: true
   feature_selection_strategy: extreme
-  n_top_features: 5
 umap:
   enabled: true
   n_neighbors: 15
--- a/configs/active/viz/alfalfa_gwas_w1w2_combined.yaml
+++ b/configs/active/viz/alfalfa_gwas_w1w2_combined.yaml
@@ -20,7 +20,6 @@ pca:
   n_components: 0.75
   standardize: true
   feature_selection_strategy: "extreme"
-  n_top_features: 5
 
 umap:
   enabled: true
--- a/configs/active/viz/alfalfa_gwas_wave1.yaml
+++ b/configs/active/viz/alfalfa_gwas_wave1.yaml
@@ -28,7 +28,6 @@ pca:
   n_components: 0.75
   standardize: true
   feature_selection_strategy: "extreme"
-  n_top_features: 5
 
 umap:
   enabled: true
--- a/configs/active/viz/alfalfa_gwas_wave1_canola.yaml
+++ b/configs/active/viz/alfalfa_gwas_wave1_canola.yaml
@@ -28,7 +28,6 @@ pca:
   n_components: 0.75
   standardize: true
   feature_selection_strategy: "extreme"
-  n_top_features: 5
 
 umap:
   enabled: true
--- a/configs/active/viz/alfalfa_gwas_wave1_canola_models.yaml
+++ b/configs/active/viz/alfalfa_gwas_wave1_canola_models.yaml
@@ -55,14 +55,14 @@ pca:
   # "top_variance" = select traits with highest total variance contribution
   feature_selection_strategy: "extreme"
 
-  # Number of traits to select using the strategy above, for PCAAnalysisStep's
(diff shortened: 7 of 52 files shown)


eberrigan (commented):
> **Verdict: COMMENT** (all findings below were fixed in commits 414f0f6 and 89ae2ce before this comment was posted)

## Review Summary

Five parallel adversarial reviews (code quality, testing/TDD, statistical rigor, performance/cross-platform, behavioral correctness) covering the diff against `main`. Core logic (the PC-scoping fix, `extreme`'s hardcoded selection, the new `select_n_features_by_variance()` helper, and the new config validation) is correct and well-tested — but the behavioral-correctness pass found 3 real numeric-edge-case bugs, and the code-quality pass found a real completeness gap in the config cleanup. All have been fixed and pushed (see commits below); this comment reflects the state **after** those fixes.

## Blocking Issues (found, then fixed — commit `414f0f6`)

1. **`n_top_features = NaN` silently bypassed both new range checks** (`NaN < 1` and `NaN >= 1` are both `False`), so it reached `PCAAnalysisStep` and crashed *after* the PCA fit with an uninformative `ValueError: cannot convert float NaN to integer`, defeating the point of validating at config-load time. Fixed with an explicit `math.isfinite()` guard.
2. **`n_top_features = inf` crashed `validate_qc_config`/`validate_viz_config` themselves with `OverflowError`**, not the `ValueError` every other check in the function raises — a caller catching `ValueError` around config validation wouldn't catch it. Fixed by the same finite-guard, checked before any arithmetic.
3. **Validation's `round()`-based whole-number tolerance disagreed with the consumption site's `int()` truncation.** A value validation certified as "whole" (e.g. `9.999999999999998` from ordinary float accumulation, within `1e-9` of `10`) still got truncated to `9` by `int()` in `PCAAnalysisStep`, silently reproducing the exact class of bug (`int(5.7)==5`) this validation exists to prevent. Fixed by using `round()` at the consumption site too, with a comment explaining why.

All three verified by reproducing them against the actual code before fixing, and each now has a regression test (`tests/test_pipeline_config.py`, `tests/test_viz_pipeline_config.py`, `tests/test_step_pca_analysis.py`).

## Important Issues (found, then fixed — commit `89ae2ce`)

- The original config cleanup covered `configs/active/viz/*.yaml` and `configs/examples/*.yaml` but missed **`configs/templates/viz_template_{with,no}_images.yaml`** — the actual scaffolding templates `/configure-run-all` copies from — and **3 root-level pre-reorg duplicate configs**, two of which (`viz_field_2024_clean.yaml`, `viz_root_coring.yaml`) `docs/troubleshooting/REPLICATION_GUIDE.md` explicitly instructs users to edit and run. Both were still claiming `n_top_features` controls UMAP-coloring count (already false pre-#206) and/or biplot arrow count. Fixed.
- `PCAConfig.n_top_features`'s docstring didn't mention the whole-number constraint applies to `top_variance`'s `>= 1` branch, only to `top_absolute`/`top_contribution`. Fixed.
- The CHANGELOG's `#206` entry undersold the blast radius — it only mentioned the 28 configs whose `n_top_features` line was removed, when actually **all 49** active configs using `"extreme"` get different `top_features.csv` output (the PC-scoping fix applies regardless of whether `n_top_features` was set). Fixed.
- Two error-message assertions checked only exception type, not the spec-required message content. Strengthened.
- One branch of `select_n_features_by_variance()` (`else: n = n_total`, unreachable via the real call site since it always passes `threshold < 1`) was uncovered by tests. Added a direct unit test.

## Suggestions (not acted on — noted for the record)

- A shared low-level `_first_index_crossing_threshold()` helper between `select_n_features_by_variance()` and `select_n_components()` would remove a narrow drift risk between their two independently-implemented "find where cumulative crosses a threshold" loops — deliberately not extracted per `design.md`'s Decision 2 (touching an existing, tested, multi-caller function for a two-caller DRY win is disproportionate; revisit if a third use case appears).
- A named constant for the `1e-9` whole-number tolerance (currently a bare literal duplicated in both `validate_qc_config`/`validate_viz_config`) would remove the "did both copies get updated together" question for free — cosmetic, not done.
- Independent re-derivation of the same variance-contribution formula in `select_n_features_by_variance()`'s caller (`perform_pca_analysis`, vectorized) vs. `select_top_features_from_pca`'s `top_variance` branch (Python loop) is mathematically identical but not guaranteed bit-identical — theoretically could disagree by one feature at an exact tie boundary. Negligible in practice (~1e-15), flagged for awareness only.
- One leftover double-blank-line from an `n_top_features` removal (`configs/active/viz/suyash_arabidopsis_pgm1_pac_2026_05_22.yaml`) — cosmetic, YAML still valid and passes its own regression test.

## Confirmed non-issues (adversarially checked, held up)

- Sort-order invariant `select_n_features_by_variance()` depends on (`feature_contributions` always sorted descending by `total_contribution`) holds unconditionally — verified by reading `perform_pca_analysis()`'s construction, not just trusting the docstring.
- Zero-variance-trait filtering and `pc_indices` scoping are orthogonal (row axis vs. column axis) — no misindexing risk.
- `top_features.csv` schema/column ordering unchanged; `generate_summary_viz.py`'s consumption of `top_features` metadata has no fixed-length assumption that the new variable-length "extreme" output would break.
- Statelessness: `PCAAnalysisStep.execute()` doesn't mutate `config` or `prev_result.metadata` in place.
- All 28 edited configs cross-checked byte-for-byte against `git diff --name-only` — exact match, no omissions.
- Cross-platform: no path-separator issues; CI green on all 3 OSes for both the fast and slow test jobs.
- Performance: diff is far from any hot path; no vectorization or memory concerns.

---
*Review by Claude Code subagent team (Code Quality · Testing · Statistical Rigor · Performance/Memory · Behavioural Correctness) — this is the PR author's own review, posted as a comment since GitHub does not allow self-approval/self-request-changes.*

PR #223, opened 2026-08-05
Fix #209: honor feature_selection in interactive PCA loading arrows

eberrigan:
## Summary

Resolves the last open item from the PCA feature-selection consolidation effort (#202, #203, #204, #206, #207 — all merged).

- `create_interactive_pca_with_images` and `create_interactive_pca_plot` (`interactive_visualization.py`) previously hardcoded `feature_contributions.nlargest(n_loadings, "total_contribution")` to pick loading-arrow features, bypassing the shared `select_top_features_from_pca()`.
- The issue asked whoever picked this up to check for real usage before deciding between "delegate" or "document as intentionally simple." That check found real, current usage: 14 of the most recent `trait_viz_*.ipynb` notebooks and `docs/PCA.md`'s Example 4 all call `create_interactive_pca_with_images(..., show_loadings=True)`. So this resolves via delegation, mirroring `create_pca_biplot`'s existing pattern.
- Both functions gain a `feature_selection: str = "top_variance"` parameter, validated unconditionally against `VALID_SELECTION_METHODS`, delegating to `select_top_features_from_pca()`. Default `"top_variance"` matches the old hardcoded ranking's typical output (both derive from the same total-variance-contribution basis); see the OpenSpec proposal for the precise (narrow) NaN/exact-tie caveat on that equivalence claim.
- Unlike `create_pca_biplot` (1-indexed `pc_x`/`pc_y`, requires a `-1` conversion), these two functions' `components` tuple is already 0-indexed and is passed through unmodified — this is pinned down with a dedicated regression test, since the same off-by-one bug class was already introduced and fixed twice elsewhere in this effort (#203, #207).
- **Bonus fix**: found and fixed a pre-existing bug in `select_top_features_from_pca()`'s `"extreme"` method (`pca.py`), newly reachable through this parameter — `n_features_to_select=0` returned every feature instead of none, due to Python's `arr[-0:] == arr[0:]` negative-slice quirk.

Full history and the 5-agent adversarial review findings (spec quality, TDD, implementation correctness, docs, git workflow — proposal review; code quality, testing, statistical rigor, performance, behavioural correctness — pre-PR self-review) are in `openspec/changes/fix-interactive-pca-feature-selection/`.

## Test plan

- [x] TDD: failing tests written first for both functions (default-equivalence, all 5 selection methods via exact-set-equality against independent `select_top_features_from_pca()` calls, PC-indexing regression, unconditional validation, `n_loadings` boundaries), then minimal implementation, then green.
- [x] `create_umap_colored_by_top_traits`'s `top_variance` branch confirmed unchanged (no code touched there).
- [x] `docs/CHANGELOG.md` `[Unreleased]` entry added.
- [x] `openspec validate fix-interactive-pca-feature-selection --strict` passes.
- [x] `uv run black --check` / `uv run ruff check` clean on all touched files.
- [x] Full suite: `uv run pytest tests/` — 3050 passed, 37 skipped, 0 failed.
- [x] Targeted suite (`test_pca.py`, `test_interactive_visualization.py`, `test_visualization.py`) — 404 passed.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -113,6 +113,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   (PR #193 review).
 
 ### Fixed
+- `create_interactive_pca_with_images` and `create_interactive_pca_plot` now
+  accept a `feature_selection` parameter and delegate their loading-arrow
+  feature selection to `select_top_features_from_pca()` (#209), instead of
+  hardcoding `feature_contributions.nlargest(n_loadings, "total_contribution")`
+  with no strategy awareness — the last unresolved consumer from the
+  `select_top_features_from_pca()` consolidation effort (#202/#203/#204/#206/
+  #207). Default `feature_selection="top_variance"` is numerically equivalent
+  to the prior hardcoded ranking (both derive from the same total-variance-
+  contribution basis), so existing callers — including the interactive-PCA
+  step of every `trait_viz_*.ipynb` analysis notebook — get identical output
+  without passing the new parameter. Unlike `create_pca_biplot`, whose public
+  `pc_x`/`pc_y` are 1-indexed and require a `-1` conversion before use as
+  `pc_indices`, these two functions' `components` tuple is already 0-indexed
+  and is passed through unmodified.
 - `PCAAnalysisStep` now always passes `pc_indices=list(range(n_components))`
   explicitly to `select_top_features_from_pca()` (#203), instead of relying
   on that function's `[0, 1]` default. Runs configured with `pca.n_components`
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -119,14 +119,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   hardcoding `feature_contributions.nlargest(n_loadings, "total_contribution")`
   with no strategy awareness — the last unresolved consumer from the
   `select_top_features_from_pca()` consolidation effort (#202/#203/#204/#206/
-  #207). Default `feature_selection="top_variance"` is numerically equivalent
-  to the prior hardcoded ranking (both derive from the same total-variance-
-  contribution basis), so existing callers — including the interactive-PCA
-  step of every `trait_viz_*.ipynb` analysis notebook — get identical output
-  without passing the new parameter. Unlike `create_pca_biplot`, whose public
-  `pc_x`/`pc_y` are 1-indexed and require a `-1` conversion before use as
-  `pc_indices`, these two functions' `components` tuple is already 0-indexed
-  and is passed through unmodified.
+  #207). Default `feature_selection="top_variance"` ranks by the same
+  total-variance-contribution basis as the prior hardcoded ranking, so
+  existing callers — including the interactive-PCA step of every
+  `trait_viz_*.ipynb` analysis notebook — get the same selected features in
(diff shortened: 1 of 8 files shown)


eberrigan (commented):
> **Verdict: APPROVE** (posted as comment — cannot approve your own PR)

## Review Summary

Clean, narrowly-scoped delegation fix: `create_interactive_pca_with_images`/`create_interactive_pca_plot` now route loading-arrow feature selection through the shared `select_top_features_from_pca()` instead of a hardcoded `nlargest` ranking, mirroring `create_pca_biplot`'s existing pattern. Along the way it catches and fixes a real, independently-verified latent bug in `select_top_features_from_pca()`'s `"extreme"` method (`n_features_to_select=0` returned every feature instead of none, due to Python's `arr[-0:] == arr[0:]` slicing quirk). All 5 reviewers (code quality, testing, statistical rigor, performance/cross-platform, behavioural correctness) found **no BLOCKING issues**. CI is green across all 13 checks (Lint, mypy baseline, Tests + Slow tests on Ubuntu/Windows/macOS, numerical-stability/reproducibility/serialization gates).

## Blocking Issues

None.

## Important Issues

None that affect correctness. Two reviewers (Testing, Behavioural Correctness) independently flagged the same minor coverage gap:
- The "`n_loadings` exceeds total feature count" tests (`test_interactive_visualization.py`, both `with_images`/`plot` variants) only exercise the default `top_variance` method, unlike the sibling `zero_loadings` tests which are parametrized across all 5 `VALID_SELECTION_METHODS`. The overflow boundary is unverified specifically for `"extreme"`, where the negative-slice logic (the exact code this PR touched) is most fragile — though tracing the logic confirms numpy's slice clamping makes it safe regardless.

## Suggestions

- (Code Quality) `create_pca_biplot` (`visualization.py:2283`) calls the same `select_top_features_from_pca()` with no guard against `top_n_features=0`, so it had the identical latent `"extreme"` bug and is silently fixed by this PR too — worth a one-line mention in the CHANGELOG/proposal so a future reader doesn't assume `create_pca_biplot` was unaffected. An easy regression test for `create_pca_biplot(feature_selection="extreme", top_n_features=0)` would be cheap insurance.
- (Testing) TDD red→green isn't independently verifiable from commit history (tests+impl landed together per commit) — self-reported via `tasks.md` only. Not disqualifying given the test specificity is otherwise excellent (independently-computed ground truth, exact-set-equality assertions), just an auditability note.
- (Testing) Consider documenting in the changelog/tests that `pca_viz_results`/`sample_pca_data` fixtures are seeded and use continuous values — that's *why* the `np.argsort` vs. `nlargest` tie-break divergence can't flake CI, but it's currently implicit.

## Verified Clean

- **Statistical rigor**: the "top_variance" equivalence claim is accurate and honestly qualified (NaN/exact-tie divergence only); no persisted output (CSV/JSON/StepResult) touched, only in-memory `go.Figure` annotations; the `n=0` fix is provably isolated to the one affected branch, with all other `n≥1` behavior byte-identical to before.
- **Performance/cross-platform**: fully vectorized (`np.argsort`), no unnecessary copies, no path/OS-specific code, and all new test assertions use set equality (not order), so `np.argsort`'s non-stable sort can't cause CI flakiness across the 3-OS matrix.
- **Behavioural correctness**: all 4 OpenSpec scenarios (default-equivalence, alternate-method exact match, unmodified `pc_indices`, unconditional validation) verified against dedicated tests; no statefulness/mutation issues.

---

